PR #1, opened 2015-10-09
Added a way to format file sizes

jstefaniuk-d2l:


test/file-size/format-file-size.js
```diff
@@ -0,0 +1,38 @@
...
+let chai = require('chai'),
+	expect = chai.expect;
+
+chai.should();
+
+describe('file-size', () => {
+	describe('format-file-size', () => {
+		[
+			{byteNum:0, expect:'0 bytes'},
+			{byteNum:-1, expect:'-1 bytes'},
```
CodeBaboon: Feels like negative sizes should return something like 'invalid' (not that term, but you see where I'm going).

jstefaniuk-d2l: Ok, I'll make it give some kind of invalid response for negatives

jstefaniuk-d2l: Actually shelved till Monday, the current ideas are keep it like it is, or on a negative show '0 bytes'

omsmith: Negatives perhaps seem fine, if you were showing a difference or something. I'm not sure we should worry about it in our context.

src/file-size/format-file-size.js
```diff
@@ -0,0 +1,35 @@
+import formatDecimal from '../number/format-decimal';
+import validateFormatValue from '../number/validate-format-value';
+
+export default function(numBytes, localeData, options) {
+	numBytes = validateFormatValue(numBytes);
+
+	const fileSizeUnits = Object.freeze([
```
omsmith: These units would be better defined outside the function

src/file-size/format-file-size.js
```diff
@@ -0,0 +1,35 @@
...
+		{unit: localeData.fileSize.units.gigabyte, num: 1073741824},
+		{unit: localeData.fileSize.units.megabyte, num: 1048576},
```
omsmith: Would prefer something like 1024 * 1024

src/file-size/format-file-size.js
```diff
@@ -0,0 +1,35 @@
...
+		{unit: localeData.fileSize.units.kilobyte, num: 1024},
+		{unit: localeData.fileSize.units.bytes, num: 1}
+	]);
+
+	let formatUnit,
+		size,
+		decimalPlaces = isNaN(options.decimalPlaces)? 2 : options.decimalPlaces;
```
omsmith: Name is a little misleading, I believe. Given it only speaks to `maximumFractionDigits`. As is, I would expect this to be passed to min and max

omsmith:
Planning to bring in locale-data for other langs?

src/file-size/format.js
```diff
@@ -0,0 +1,15 @@
...
+import localeProvider from '../locale-provider';
+
+export default class FileSizeFormat {
+	constructor(locale, options) {
+		this.options = options || {};
+		this.localeData = localeProvider(locale, options.locale);
+	}
+	formatFileSize(fileSize, options){
+		for(var option in options){
+			this.options[option] = options[option];
```
omsmith: You're overwriting options of the instance here.

```js
const format = new FileSizeFormat('en', { decimalPlaces: 2 });

format.formatFileSize(1234); // 1.23 KB
format.formatFileSize(1234, { decimalPlaces: 3 }); // 1.234 KB
format.formatFileSize(1234); // 1.234 KB
```

src/file-size/format.js
```diff
@@ -0,0 +1,15 @@
+import FileSizeFormatter from './format-file-size';
+import localeProvider from '../locale-provider';
+
+export default class FileSizeFormat {
+	constructor(locale, options) {
+		this.options = options || {};
+		this.localeData = localeProvider(locale, options.locale);
+	}
+	formatFileSize(fileSize, options){
+		for(var option in options){
```
omsmith: What if `options` isn't provided?

test/file-size/format-file-size.js
dbatiste: My thoughts exactly.  Negative should be fine.

jstefaniuk-d2l:
@omsmith I think I read on stash that if you add to one of the locale-data files that they get sent off and auto-updated at the end of the month

CodeBaboon:
@jstefaniuk-d2l @omsmith 
Not sure how to link to a Slack comment but this is what Jeff is referring to (from the General channel):

mike_ruthven [12:23 PM] 
@channel: Wanted to communicate a process for getting string translations done. Currently, I am handling iOS and Android app string translations on a monthly basis by sending the strings out. If the changes you are making don’t add new files to this process, then anything you are doing will get translated “magically”. @surekha has been handling the translations for the services, where they are needed. If you are adding a completely new file or library for iOS or Android, please talk to me about making it part of the standard translation work, and speak to Surekha when you have new strings for one of the services. Between us we will get the translations looked after.

README.md
```diff
@@ -210,6 +210,35 @@ console.log(
...
+
+Syntax:
+
+```javascript
+var formatFS = new d2lIntl.FileSizeFormat(locale[, options]);
+```
+Options:
+
+- **locale**: see [overriding locale data](#Overriding Locale Data)
+- **decimalPlaces**: Change the default maximum amount of decimal places to include in the result (Default: 2)
```
jstefaniuk-d2l: I'm going to update the decimalPlaces readme text to maximumFractionDigits in the next PR

dbatiste:
@jstefaniuk-d2l : that process is currently limited to Stash (and I am not sure which repos it actually includes).  Currently, translations for FRAs and other components on GitHub must be handled manually.  The process for collecting, sending, and processing translations is currently being over-hauled to include identified (via some type of registration) from both GitHub and Stash. @njostonehouse / @BoGs : not sure if you want to add anything to that?

omsmith:
@jstefaniuk-d2l You can find the relevant langterms which you'll want to port the formats over from in FileSizePlainTextRenderer (use code search to find it). They're all very simple - slight variations on en; no unicode characters or anything as far as I saw.

jstefaniuk-d2l:
@omsmith Thanks!

omsmith:
As evidenced by FileSizePlainTextRenderer, you should have a special case for `1`

dlockhart:
Yeah no need to send anything off for translation here, as Owen said we should just copy in all the existing term values.

src/file-size/format.js
```diff
@@ -0,0 +1,20 @@
+import FileSizeFormatter from './format-file-size';
+import localeProvider from '../locale-provider';
+
+export default class FileSizeFormat {
+	constructor(locale, options) {
+		this.options = options || {};
+		this.localeData = localeProvider(locale, options.locale);
+	}
+	formatFileSize(fileSize, options){
+		options = options || {};
```
dlockhart: I don't think we need `options` here (yet). Otherwise people will override the number of decimals _because they can_ and we'll end up with inconsistencies across UIs. Better to just pick a good default (same as what we do in the LMS) and only add an override later if it's really needed.

src/file-size/format.js
```diff
@@ -0,0 +1,20 @@
+import FileSizeFormatter from './format-file-size';
+import localeProvider from '../locale-provider';
+
+export default class FileSizeFormat {
+	constructor(locale, options) {
+		this.options = options || {};
+		this.localeData = localeProvider(locale, options.locale);
+	}
+	formatFileSize(fileSize, options){
```
dlockhart: Can this method just be called `format`? That way it's consistent with the number and date formatters.

src/locale-data/en-US.json
```diff
@@ -104,5 +104,16 @@
 			"percent": "%"
 		},
 		"groupSize": 3
+	},
+	"fileSize": {
```
dlockhart: You'll want to bring in the locale data for all the other locales as well. It should all be there contained in the language terms the existing formatter is using.

src/locale-data/en.json
```diff
@@ -104,5 +104,16 @@
...
+	},
+	"fileSize": {
+		"patterns": {
+			"fileSizePattern": "{number} {unit}"
+		},
+		"units" : {
+			"gigabyte": "GB",
+			"megabyte": "MB",
+			"kilobyte": "KB",
+			"bytes": "bytes"
```
dlockhart: You'll likely want a term for the singular "byte" (e.g. "1 byte") as well.

src/file-size/format.js
omsmith: (LMS does 2 places, without trailing zeroes)

README.md
```diff
@@ -210,6 +210,35 @@ console.log(
...
+
+Syntax:
+
+```javascript
+var formatFS = new d2lIntl.FileSizeFormat(locale[, options]);
+```
+Options:
+
+- **locale**: see [overriding locale data](#Overriding Locale Data)
+- **maximumFractionDigits**: Change the default maximum amount of decimal places to include in the result (Default: 2)
```
dlockhart: As noted below, I don't think this option is necessary, let's just use a good default and add it later if someone really needs it.

omsmith:
All the `options` params in the added functions should be removed at this point, no?

README.md
```diff
@@ -210,6 +210,30 @@ console.log(
...
+The `FileSizeFormat` object can be used to format a file size appropriately for the user's locale.
+
+Syntax:
+
+```javascript
+var formatFS = new d2lIntl.FileSizeFormat(locale[, options]);
+```
+Options:
+
+- **locale**: see [overriding locale data](#Overriding Locale Data)
```
omsmith: This isn't an option (member of the `options` object).

README.md
```diff
@@ -210,6 +210,30 @@ console.log(
...
+
+To format a file size, call the `formatFileSize` method:
```
omsmith: function name needs updating

src/locale-data/ko-KR.json
```diff
@@ -104,5 +104,17 @@
 			"percent": "%"
 		},
 		"groupSize": 3
+	},
+	"fileSize": {
+		"patterns": {
+			"fileSizePattern": "{number} {unit}"
```
dlockhart: According [to the old terms](http://search.dev.d2l/source/xref/LSOne/lp/_lang/LangTerms/translations/ko-kr/Standard.xml#585), I think the pattern should not have a space: `{number}{unit}`

src/locale-data/ko.json
```diff
@@ -104,5 +104,17 @@
 			"percent": "%"
 		},
 		"groupSize": 3
+	},
+	"fileSize": {
+		"patterns": {
+			"fileSizePattern": "{number} {unit}"
```
dlockhart: Same here, no space.

test/file-size/format-file-size.js
```diff
@@ -0,0 +1,35 @@
...
+			{byteNum:1023, expect:'1,023 Bytes'},
+			{byteNum:-1023, expect:'-1,023 Bytes'},
+			{byteNum:1024, expect:'1 KB'},
+			{byteNum:-1024, expect:'-1 KB'},
+			{byteNum:1048576, expect:'1 MB'},
+			{byteNum:-1048576, expect:'-1 MB'},
+			{byteNum:1073741824, expect:'1 GB'},
+			{byteNum:1073741824 * 2, expect:'2 GB'},
+			{byteNum:-1073741824, expect:'-1 GB'},
+			{byteNum:1881111114, expect:'1.75 GB'},
```
dlockhart: Could you add some tests for the other locales now that they're in there? You should just be able to pass in a locale override.

src/file-size/format.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default class FileSizeFormat {
+	constructor(locale, options) {
+		this.options = options || {};
+		this.localeData = localeProvider(locale, options.locale);
+	}
+	format(fileSize, options){
+		options = options || {};
+
+		for(var option in this.options) {
```
dlockhart: Now that there aren't any `options` anymore, this can go away.

test/file-size/format-file-size.js
```diff
@@ -0,0 +1,35 @@
...
+		].forEach(input => {
+			it(`should parse the file-size: "${input.byteNum}" bytes`, () => {
+				var options;
+				var formatFileSize = new format('en', {});
+				const parsedFileSize = formatFileSize.format(input.byteNum, options);
```
dlockhart: No need for the `options` variable or to pass it in.

src/locale-provider.js
```diff
@@ -13,6 +13,7 @@ const allLocales = {
 	'fr': require('./locale-data/fr.json'),
 	'ja': require('./locale-data/ja.json'),
 	'ko-KR': require('./locale-data/ko-KR.json'),
+	'ko': require('./locale-data/ko.json'),
```
dlockhart: Good catch :)

src/file-size/format-file-size.js
```diff
@@ -0,0 +1,43 @@
...
+			if (Math.abs(numBytes) >= unitSize.num) {
+				formatUnit = unitSize.unit;
+				size = formatDecimal(numBytes/unitSize.num, localeData.number, { maximumFractionDigits: 2 });
+				break;
+			}
+		}
+	}
+
+	let ret = localeData.fileSize.patterns.fileSizePattern;
+	ret = ret.replace('{number}', size).replace('{unit}', formatUnit);
```
dlockhart: @jstefaniuk-d2l: could you use [process-pattern](https://github.com/Brightspace/intl/blob/master/src/util/process-pattern.js) here instead of the double replace? The other formatters should have examples of how to use it.

PR #4, opened 2015-10-16
conversion to ES5 from ES6

dlockhart:
@dbatiste, @omsmith: this removes the need to have a transpiled version in "lib" and will allow us to publish to Bower without needing to check in the transpiled version.


omsmith:
Seems like a pretty straightforward port

PR #9, opened 2016-05-03
Add support for variable number of group sizes for format_decimal

akwiatkowskid2l:


src/number/format-decimal.js
```diff
@@ -42,7 +51,11 @@ module.exports = function(value, localeData, options) {
 
 		digitNum++;
 
-		if (digitNum > 1 && digitNum % localeData.groupSize === 1) {
+		if (digitNum > 1 && digitNum % currentGroupSize === 1) {
+			if ( groupSizes.length > 0) {
+				currentGroupSize = groupSizes.shift();
+			}
+			digitNum = 1;	// reset this back to 1 to handle variable group sizes
```
omsmith: The initial value was `0`

src/number/format-decimal.js
```diff
@@ -19,6 +19,15 @@ module.exports = function(value, localeData, options) {
...
 	var digitNum = 0;
+	var currentGroupSize ;
+	var groupSizes;
+	if ( Array.isArray( localeData.groupSize) ) {
+		groupSizes = localeData.groupSize.slice(); 	// make a copy of the groupsizes
+		currentGroupSize = groupSizes.shift();		// pop the first groupSize from the front
+	} else {
+		groupSizes = [];
+		currentGroupSize = localeData.groupSize;
+	}
```
dlockhart: Might be cleaner to just default to `[localeData.groupSize]`? e.g. these 8 lines could become:

    var groupSizes = Array.isArray(localeData.groupSizes) ?
      localeData.groupSizes.slice() : [localeData.groupSizes];
    var currentGroupSize = groupSizes.shift();

test/number/format-decimal.js
```diff
@@ -152,6 +152,19 @@ describe('NumberFormat', function() {
...
+				var value = format(1000000.01, localeData, options);
+				expect(value).to.equal('1000000.01');
+			});
+
+			it('should handle variable group sizes', function() {
+				localeData.groupSize = [ 3, 2, 0 ];
+				var options = {maximumFractionDigits: 2};
+				var value = format(1000000.01, localeData, options);
+				expect(value).to.equal('10,00,000.01');
+			});
```
dlockhart: Based on [Microsoft's documentation](https://msdn.microsoft.com/en-us/library/system.globalization.numberformatinfo.numbergroupsizes(v=vs.110).aspx), it should use the last group size value indefinitely. So maybe a test case for that? e.g. groupSize = [4,2], input = "123456789.01", output = "1,23,45,6789.01"

test/number/format-decimal.js
```diff
@@ -152,6 +152,19 @@ describe('NumberFormat', function() {
...
 			});
 
+			it('should handle group size of 0', function() {
+				localeData.groupSize = 0;
+				var options = {maximumFractionDigits: 2};
+				var value = format(1000000.01, localeData, options);
+				expect(value).to.equal('1000000.01');
+			});
+
+			it('should handle variable group sizes', function() {
```
dlockhart: I'm not sure this test case it really testing the behaviour of the "0" value, which should stop further grouping from occurring. A more thorough input/output might be: input = "1000000000.01", output = "10000,00,000.01"

test/number/format-decimal.js
akwiatkowskid2l: Agreed. Will put that in.

test/number/format-decimal.js
akwiatkowskid2l: This was intended to test the case with no group separators at all.

src/number/format-decimal.js
dlockhart: Instead of `slice()` to copy and `shift()` to pop, could you just move a pointer around within the existing array? @omsmith and I are worried about the unnecessary performance hits of this array manipulation if there are many numbers being rendered on the page.

src/number/format-decimal.js
akwiatkowskid2l: This caught me off guard as well. But because line 54 has `digiNum % currentGroupSize ===1`, this 'counter' should start at 1 as well. For example, for group sizes of 3, you want to the code to run on the 4th,7th, and 10th digits.
If we set digitNum=0, it would run on the 4th, 8th and 12th. 
Would it be more readable if I changed `digitNum = 1` to `digitNum = digitNum - currentGroupSize`?

src/number/format-decimal.js
akwiatkowskid2l: Sure no problem. I used that solution originally, but I thought this was more succinct.  I am by no means a Javascript expert however, and I defer to your guys' judgment.

src/number/format-decimal.js
omsmith: Perhaps? The word _reset_ here is what made me notice, and I would assume it was a bug if I saw it ("oops, they accidently 'reset' it to the wrong value"). So yeah, either needs to be explained differently, or the arrangement should change. What happens when the initial value of `digitNum` is 1 instead of `0`?

akwiatkowskid2l: If you started with digitNum=1, then the counting would be off by one. So you'd get 12,345,67, instead of 1,234,567. 
The algorithm goes right to left, so you want to add commas when digitnum=4 (in this case '4'), and when digitnum=7 (in this case '1')

The confusion comes because now that we have variable group sizes. I could re-factor this to use countdown group counters instead. That might be easier to understand. Let me give that a try and see what you guys think.

omsmith:
Much easier to grok, thanks :)!

PR #11, opened 2016-05-06
number: format numbers faster

omsmith:


omsmith:
~2x faster across all browsers

omsmith:
Removed the code-path without coverage, since it's can't actually occur it turns out

dlockhart:
Neat! I like 2x faster!

PR #12, opened 2016-05-10
Update fr.json

Spidersouris:
https://en.wikipedia.org/wiki/Date_format_by_country


omsmith:
Hi @Spidersouris, thanks for the contribution.

This is however consistent with how dates have been formatted in our system for a long time - so we'll have to stick with the status quo for now. May look into this in the future though, and will revisit it then.

Thanks again :)

PR #13, opened 2016-06-08
Jtarka/decimal format/large and small numbers

JasonTarka-D2L:
**Fix formatting of really small decimals**

Small decimals, expressed in scientific/exponential notation, weren't being handled correctly and would result in totally broken results.
eg: `1e-9` would become `0.-9`
This change will put the appropriate number of zeros before the significant digits when scientific notation is detected.

**Handle numbers greater than 32-bit integers**

OR-ing the number against `0` would OR it against a 32-bit integer, causing issues with numbers longer than 32-bits.
eg: `10000000000 | 0 == 1410065408`
`Math.trunc()` simply drops the decimal from a number, which appeared to be the intention of OR-ing with 0.

### DIFF
--- a/src/number/format-decimal.js
+++ b/src/number/format-decimal.js
@@ -22,10 +22,16 @@ module.exports = function(value, localeData, options) {
 	var decimalStr = null;
 
 	if (hasDecimal) {
-		// get a string of 0.xxx
-		decimalStr = '' + (Math.round((value - integerValue) * precisionScaling) / precisionScaling);
-		// the first decimal place is index 2
-		decimalStr = decimalStr.slice(2);
+		var decimalValue = Math.round( (value - integerValue) * precisionScaling) / precisionScaling;
+		if (decimalValue.toExponential() === decimalValue.toString()) {
+			// Get a string with leading zeros
+			decimalStr = formatExponentialDecimal(decimalValue);
+		} else {
+			// get a string of 0.xxx
+			decimalStr = '' + decimalValue;
+			// the first decimal place is index 2
+			decimalStr = decimalStr.slice(2);
+		}
 	} else if (options.minimumFractionDigits > 0) {
 		decimalStr = '';
 	}
@@ -48,3 +54,17 @@ module.exports = function(value, localeData, options) {
 	return ret;
 
 };
+
+function formatExponentialDecimal( value) {
+	// Get a value like "1.25e-8" and find its values
+	value = parseFloat(value).toExponential();
+
+	var pieces = value.split( 'e-' ),
+		ret = pieces[0].replace( pieces[0].charAt(1), '' ),
+		zeroCount = parseInt( pieces[1] ) - 1;
+
+	for (var i = 0; i < zeroCount; i++) {
+		ret = '0' + ret;
+	}
+	return ret;
+}
--- a/src/number/format-decimal.js
+++ b/src/number/format-decimal.js
@@ -14,7 +14,7 @@ module.exports = function formatDecimal(value, localeData, options) {
 	// round to desired precision
 	value = Math.abs(Math.round(value * precisionScaling) / precisionScaling);
 
-	var integerValue = Math.trunc( value );
+	var integerValue = Math.floor(value);
 
 	var ret = formatPositiveInteger(integerValue, localeData, options);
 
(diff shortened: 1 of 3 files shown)


omsmith:
Thanks @JasonTarka-D2L, I guess I underestimated how quickly the 32-bit limit comes up in my head :)

src/number/format-decimal.js
```diff
@@ -14,18 +14,24 @@ module.exports = function(value, localeData, options) {
...
 	var hasDecimal = value !== integerValue;
 	var decimalStr = null;
 
 	if (hasDecimal) {
-		// get a string of 0.xxx
-		decimalStr = '' + (Math.round((value - integerValue) * precisionScaling) / precisionScaling);
-		// the first decimal place is index 2
-		decimalStr = decimalStr.slice(2);
+		var decimalValue = Math.round( (value - integerValue) * precisionScaling) / precisionScaling;
+		if (decimalValue.toExponential() === decimalValue.toString()) {
```
omsmith: We're `toString`ing multiple times now. Could this go back to `decimalStr = '' + (...)` and then `if (decimalStr === decimalValue.toExponential()) { formatExponentialDecimalString(decimalStr); }`

omsmith:
Question, the crossover point on my machine is _1e-7_ (0.000001). Are we actually rendering that precision in our UIs somewhere?

src/number/format-decimal.js
```diff
@@ -48,3 +54,17 @@ module.exports = function(value, localeData, options) {
 	return ret;
 
 };
+
+function formatExponentialDecimal( value) {
+	// Get a value like "1.25e-8" and find its values
+	value = parseFloat(value).toExponential();
+
+	var pieces = value.split( 'e-' ),
+		ret = pieces[0].replace( pieces[0].charAt(1), '' ),
```
omsmith: Should this just be `('.', '')`?

src/number/format-decimal.js
```diff
@@ -14,18 +14,24 @@ module.exports = function(value, localeData, options) {
 	// round to desired precision
 	value = Math.abs(Math.round(value * precisionScaling) / precisionScaling);
 
-	var integerValue = value | 0;
+	var integerValue = Math.trunc( value );
```
omsmith: I think you'll need the polyfill for this (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc)

omsmith:
Style note, which I guess isn't in our eslint ruleset as is, we don't put spacing in the parentheses.

dbatiste:
Looks like eslint config version is out of date.  Still might need another rule for that spacing though.  :(

src/number/format-decimal.js
JasonTarka-D2L: Good catch. Will change in my next commit.

src/number/format-decimal.js
JasonTarka-D2L: Was thinking some locales might use a comma instead, but checking the [ES5 spec][0], it specifies a period. Will change.

[0]: http://www.ecma-international.org/ecma-262/5.1/#sec-7.8.3

src/number/format-decimal.js
JasonTarka-D2L: Or use `Math.floor()`, which is supported everywhere. This is forced to be a positive value anyway via `Math.abs()`, so negative rounding cases aren't a concern.

omsmith: Yep, should be fine.

JasonTarka-D2L:
Always seems kinda weird to me that our JS projects follow a different style than [our C# style guide][style], for exactly the same constructs.

As for where do we show numbers that small, I'm not aware of any clients having this sort of issue, but we've produced it using existing features, albeit in a forced fashion. But it came from running into clients using partial points in rubrics (eg: level is worth 0.2, another 0.4, etc.) causing floating point issues, so it wouldn't surprise me if a user does something to cause this.

[style]: http://docs.dev.d2l/index.php/Coding_Standards_-_CSharp#Statements

omsmith:
Alrighty. It's not strictly about how small the numbers get really - it's the large `maximumFractionDigits` that I'm surprised about.

omsmith:
LGTM

omsmith:
I'll get this in tomorrow

PR #16, opened 2017-07-07
Add the ability to retrieve the locale data directly

alogan519:
Add an import point for html


d2l-intl.html
```diff
@@ -0,0 +1 @@
+<script src="./dist/Intl.js"></script>
```
alogan519: This gives us an easy way to import into Polymer elements

README.md
```diff
@@ -26,6 +26,12 @@ Or install from [NPM]([npm-url]):
 npm install d2l-intl
 ```
 
+Or include from html:
```
dlockhart: We should probably be a bit more specific here, since `rel="import"` doesn't exist in browsers other than Chrome (yet). I'd say something like: "Or, include it as a [Polymer](https://www.polymer-project.org/) HTML import:"

PR #24, opened 2018-11-01
Converting from CommonJS to ES6 modules

dlockhart:
This converts the `d2l-intl` library to be ES6-module based, as a pre-cursor to being able to use it in `localize-behavior` from Polymer 3.

I've added a build step that will transpile to ES5 and the `main` in `package.json` points to that version so that `require('d2l-intl')` calls from NPM will continue to work.

I've also removed the step that publishes to the CDN, as going forward I think we want browser-based usages to use the `import`.


dlockhart:
Thoughts, concerns, fears, doubts, hopes?

omsmith:
Thoughts: Pie can be pretty good.
Concerns: Phones are awfully distracting, I often find myself reaching for it in between everything I do. It's something I need to work on.
Fears: I dunno about fearful of, but heights and dark spaces and be disconcerting.
Doubts: Jays succeeding without Gibby
Hopes: I will continue to learn and grow at D2L, including about ES6 and Polymer 3 I guess

PR #26, opened 2019-03-21
Fuzzy date support

ctwomey1:
Adapted from https://github.com/Brightspace/d2l-activity-details-ui/blob/3d616ea18b625f7eec60cc3b0a592ad538e9d180/src/behaviors/fuzzy-date-behavior.js to implement fuzzy dates for more general consumption.

The contents of this PR allow for the following usages like:
```
> const format = require('./dist/date-time/format');
> new format('en', {format:'fuzzy'}).format(new Date())
'Just now'
> new format('en', {format:'fuzzy'}).format(new Date('March 21, 2019 7:38 am'))
'1 minute ago'
> new format('en', {format:'fuzzy'}).format(new Date('March 21, 2019 7:37 am'))
'2 minutes ago'
> new format('en', {format:'fuzzy'}).format(new Date('March 21, 2019 7:34 am'))
'5 minutes ago'
> new format('en', {format:'fuzzy'}).format(new Date('March 21, 2019 6:38 am'))
'1 hour ago'
> new format('en', {format:'fuzzy'}).format(new Date('March 21, 2019 5:38 am'))
'2 hours ago'
```

Tests run in both Chrome 72 and Firefox 66.  Still need to verify in IE 11 and Edge (Mocha doesn't seem to run in either)

### DIFF
--- a/src/date-time/format-fuzzy-date.js
+++ b/src/date-time/format-fuzzy-date.js
@@ -13,10 +13,10 @@ const oneHour = 3600000;
 const oneHourThirtyMinutes = 5400000;
 const twoHoursThirtyMinutes = 9000000;
 
-export default function FuzzyDateFormatter(localeData, locale, timeZone) {
+export default function FuzzyDateFormatter(localeData, locale, timezone) {
 	this.format = localeData.date.formats.fuzzyFormats;
 	this.locale = locale;
-	this.timeZone = timeZone ||  new Intl.DateTimeFormat().resolvedOptions().timeZone;
+	this.timezone = timezone ||  new Intl.DateTimeFormat().resolvedOptions().timeZone;
 }
 
 /*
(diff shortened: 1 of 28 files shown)


src/date-time/format-fuzzy-date.js
```diff
@@ -0,0 +1,82 @@
...
+
+const oneMinute = 60000;
+const twoAndAHalfMinutes = 150000;
+const fortyFiveMinutes = 2700000;
+
+const oneHour = 3600000;
+const oneHourThirtyMinutes = 5400000;
+const twoHoursThirtyMinutes = 9000000;
+
+export default function FuzzyDateFormatter(localeData, locale) {
```
ctwomey1: Stayed consistent with the repo here instead of jumping up to class syntax.

ctwomey1:
Seems providing a timezone in the string doesn't quite work for the "Yesterday" syntax, i'll fix that up

src/date-time/format-fuzzy-date.js
```diff
@@ -0,0 +1,85 @@
...
+const fortyFiveMinutes = 2700000;
+
+const oneHour = 3600000;
+const oneHourThirtyMinutes = 5400000;
+const twoHoursThirtyMinutes = 9000000;
+
+export default function FuzzyDateFormatter(localeData, locale, timeZone) {
+	this.format = localeData.date.formats.fuzzyFormats;
+	this.locale = locale;
+	this.timeZone = timeZone ||  new Intl.DateTimeFormat().resolvedOptions().timeZone;
```
ctwomey1: ie 11 doesn't resolve this.  However using an undefined timezone for intl.DateTimeFormat in ie 11 behaves like it uses the existing client timezone.  Ideally we'd always be passing in a configured zone

src/date-time/format-fuzzy-date.js
```diff
@@ -0,0 +1,85 @@
...
+/*
+* Returns the fuzzy date string as determined by difference between now and date
+*/
+FuzzyDateFormatter.prototype.getDateString = function(inputDate, nowDate)  {
+
+	const referenceDate = nowDate || new Date();
+	const calendarDateDiff = (referenceDate - inputDate) / msPerDay;
+	const calendarDaysFloored = Math.floor(calendarDateDiff);
+	const sameYear = referenceDate.getYear() === inputDate.getYear();
+	const configuredTimeZoneHours = parseInt(new Intl.DateTimeFormat(this.locale, {hour:'numeric', hour12:false, timeZone: this.timeZone}).format(referenceDate).split(':'));
```
ctwomey1: new Intl.DateTimeFormat('en-us', {hour:'numeric', hour12:false}).format(new Date()) === 09 in firefox and chrome
new Intl.DateTimeFormat('en-us', {hour:'numeric', hour12:false}).format(new Date()) === 9:00 in edge/ie

ctwomey1:
@dlockhart I added support for passing in a timezone.  Can you give this another quick pass?

Barring any changes I feel like this is good to squash and merge, (I don't have write access to this repo)

src/date-time/format.js
```diff
@@ -1,11 +1,13 @@
 import formatTime from './format-time.js';
 import formatDate from './format-date.js';
+import formatFuzzyDate from './format-fuzzy-date.js';
 import localeProvider from '../locale-provider.js';
 
 export default function DateTimeFormat(locale, options) {
 	options = options || {};
 	this.options = options;
 	this.localeData = localeProvider(locale, options.locale);
+	this._fuzzyDateFormatter = new formatFuzzyDate(this.localeData, this.options.locale, this.options.timeZone);
```
dlockhart: Could you use lowercase `timezone` so that it's consistent [with the existing stuff](https://github.com/Brightspace/intl/blob/master/src/date-time/format-time.js#L7)?

ctwomey1: will do!

ctwomey1: So interestingly, Intl.DateTime accepts only the timezone Identifier eg "America/Winnipeg".  Time formatting uses the abbreviation.

Do you have a preference on the nomenclature?  I was thinking `option.timezoneidentifier`

dlockhart: Sounds good to me

dlockhart:
Any update on this PR?

ctwomey1:
I was going to use this, but some requirements are changing and I won't need to render fuzzy dates anymore.

I can still merge this if you think it's valuable, but I think closing might be the best option for now until a use case presents itself

dlockhart:
Cool yeah let's close it until we know it'll actually be used.

PR #27, opened 2019-05-02
Add typescript types

awikkerink:


awikkerink:
Correct. But shouldn't be a huge problem unless there is a major version bump

awikkerink:
@dlockhart could you merge? I don't have write permission

PR #31, opened 2019-12-31
Version 3

dlockhart:
A bunch of major changes going on here, so this will be a major version bump to `v3`. I'll be going through libraries that depend on this (core, BSI, localize-behavior, date/time components) and updating them to use this version so we can upgrade everything in BSI together.

It's worth noting that although there are a lot of changes here around how code is organized, the underlying algorithms for parsing/formatting haven't changed -- other than adding a few `const`s and `let`s.

### Size and Requests

The main motivation for these changes is a reduction in the overall size of the library, as well as the ability to import only the functionality you need (e.g. just number stuff or just date stuff) and not have to download the rest. Additionally when serving things unbundled, we don't want a ton of unnecessary network requests.

**Previously:** 52 requests (always), 74 KB uncompressed, 36 KB compressed
**Now:** 4 requests (or 2 requests if you only need a subset of functionality), 26 KB uncompressed, 9 KB compressed

So it's now about 1/4 the size and a dramatic reduction in the number of requests.

The way I was able to do this was by removing all the `en.json`/`ar.json`/etc. locale files and merging all the locale descriptor data into the functions themselves. While less clean, this resulted in a massive reduction in the library size and number of requests.

### Automatic Language Resolution

Previously, the language and any user/org overrides were passed into each function. This forced consumers to resolve those on their own -- leading to inconsistencies and duplication -- or required the use of a wrapper library like `d2l-localize-behavior` or `@brightspace-ui/core/mixins/localize-mixin.js`. Now folks will be able to use `intl` on its own.

With these changes, the automatic language, override and other `data-` attribute resolution from the document are moved into this library. When `@brightspace-ui/core` adopts these changes, its copy of this code will be removed.

If required, I may add in the ability to pass in a language code as an option to the various functions, but for now it's not needed so it'll always use the automatic resolution.

### API Changes

I didn't see a need to maintain the old OOP-style API, where you had to first create a formatter or parser and then call a method on it. Instead you now just import the format/parse function(s) you need and call them directly.

Originally I think I'd hoped that by mirroring the native browser APIs we could one day move to them. We can still switch some of this code to call into the native APIs if we want, but I don't see a strong case to mirror their API.


lib/number.js
```diff
@@ -0,0 +1,318 @@
...
+	validateFormatValue
+} from './common.js';
+
+function formatPositiveInteger(value, descriptor) {
+	value = Math.floor(value);
+
+	const valueStr = '' + value;
+	let ret = '';
+
+	const groupSizes = [3];
```
dlockhart: Every locale uses a group size of `3`, so I simplified the code here a bit. I may go through and simplify it further at some point.

package.json
```diff
@@ -1,24 +1,23 @@
...
-  "name": "d2l-intl",
-  "version": "2.1.0",
-  "description": "D2L internationalization APIs for number, date and time formatting and parsing.",
-  "main": "dist/index.js",
-  "types": "src/index.d.ts",
-  "module": "src/index.js",
+  "name": "@brightspace-ui/intl",
+  "version": "3.0.0",
+  "description": "Internationalization APIs for number, date, time and file size formatting and parsing in D2L Brightspace.",
+  "main": "lib/number.js",
```
dlockhart: I didn't want to keep `index.js` around, since it would encourage folks to just import everything from there -- not just the things they needed.

package.json
```diff
@@ -1,24 +1,23 @@
...
-    "prepublish": "npm run build"
+    "test": "npm run lint -s && npm run test:unit -s"
   },
+  "files": [
+    "/lib"
+  ],
   "repository": {
     "type": "git",
-    "url": "https://github.com/Brightspace/intl.git"
+    "url": "https://github.com/BrightspaceUI/intl.git"
```
dlockhart: I'll be moving the repo once this merges.

test/common.js
```diff
@@ -0,0 +1,215 @@
...
+		htmlElem.removeAttribute('data-intl-overrides');
+		htmlElem.removeAttribute('data-timezone');
+		documentLocaleSettings.reset();
+	});
+
+	it('should default to "en"', () => {
+		expect(defaultLocale).to.equal('en');
+	});
+
+	describe('getLanguage', () => {
```
dlockhart: A lot of these tests are from core -- I'll remove them from there once it's updated to use this.

README.md
```diff
@@ -1,252 +1,173 @@
...
 
-To format a **time only** (without the date portion), use the `formatTime` method:
-
-```javascript
-var formatter = new d2lIntl.DateTimeFormat('ko');
-var time = formatter.formatTime(
-	new Date(2015, 8, 23, 14, 5)
-); // -> 오후 14:05
+const value = formatDate(
+	ew Date(2015, 8, 23),
```
margaree: nit: `ew`

margaree: and missing `'` on the line below around `full`

dlockhart: lol thank-you

lib/common.js
```diff
@@ -0,0 +1,183 @@
...
+		if (normalized === this._language) return;
+		this._language = normalized;
+		this._listeners.forEach((cb) => cb());
+	}
+
+	get fallbackLanguage() { return this._fallbackLanguage; }
+	set fallbackLanguage(val) {
+		const normalized = this._normalize(val);
+		if (normalized === this._fallbackLanguage) return;
+		this._fallbackLanguage = this._normalize(val);
```
margaree: nit: could just use `this._fallbackLanguage = normalized` like in `set language`

lib/dateTime.js
```diff
@@ -0,0 +1,580 @@
...
+			dayPeriods: {am: dayPeriods[0], pm: dayPeriods[1]}
+		}
+	};
+
+	const settings = getDocumentLocaleSettings();
+	if (settings.overrides.date) {
+		merge(descriptor, settings.overrides.date);
+	}
+
+	// TODO: fix 12/24 hour clock bug with missing info in format string
```
margaree: Is this still applicable? Should there be an issue for it?

lib/dateTime.js
```diff
@@ -0,0 +1,580 @@
...
+				['أحد', 'إثنين', 'ثلاثاء', 'أربعاء', 'خميس', 'جمعة', 'سبت'],
+				['أ', 'إ', 'ث', 'أر', 'خ', 'ج', 'س']
+			];
+			firstDayOfWeek = 6;
+			weekendStartDay = 4;
+			weekendEndDay = 5;
+			break;
+		case 'da':
+			dateFormats = ['dddd \'den\' d. MMMM yyyy', 'd. MMM. yyyy', 'dd/MM/yyyy', 'MMMM yyyy', 'd. MMMM'];
+			months = [numericMonthsLong, numericMonthsLong, numericMonthsNarrow];
```
margaree: This array and the one for months for `de` are length 3, but it seems others are length 2 with long and short versions of months. Is this expected?

lib/dateTime.js
dlockhart: Yeah good call. I'd originally thought I'd fix this existing bug as part of this work, but I think I'll do it later -- I'll create an issue.

dlockhart: Created #32.

lib/dateTime.js
dlockhart: Great catch, thanks! The [place I pulled the data from](https://github.com/andyearnshaw/Intl.js/blob/master/locale-data/json/de-DE.json#L77) for these new locales had long, short and narrow month names but we only need long and short.

lib/fileSize.js
```diff
@@ -0,0 +1,92 @@
...
+
+export function getFileSizeDescriptor() {
+
+	const language = getLanguage();
+
+	const val = {
+		'gigabyte': 'GB',
+		'megabyte': 'MB',
+		'kilobyte': 'KB',
+		'bytes': 'bytes',
```
margaree: This probably doesn't matter but just want to confirm this change is expected - it looks like previously for the `en` locales 'byte' and 'bytes' were actually capitalized (e.g., 'Bytes').

lib/fileSize.js
```diff
@@ -0,0 +1,92 @@
...
+			val.megabyte = 'Mo';
+			val.kilobyte = 'Ko';
+			val.bytes = 'octets';
+			val.byte = 'octet';
+			break;
+		case 'ja':
+			val.bytes = val.byte = 'バイト';
+			break;
+		case 'ko':
+			val.bytes = val.byte = '바이트';
```
margaree: It looks like this was previously slightly different `0]바이트` (https://github.com/Brightspace/intl/blob/master/src/locale-data/ko-KR.js#L116) so just want to confirm this is expected.

lib/fileSize.js
dlockhart: Yep, very observant! I made that as a conscious change.

lib/fileSize.js
dlockhart: Yep, it was a typo in the old term with an extra `0]`. I think when it was copied [from the LMS](http://search.d2l.dev/xref/lms/lp/_lang/LangTerms/translations/ko-kr/Standard.xml?r=0f2ceefc#787) -- whose terms had `[0]` replace strings in them -- it was accidentally left in.

lib/number.js
```diff
@@ -0,0 +1,318 @@
...
+	return value;
+
+}
+
+export function getNumberDescriptor() {
+
+	const language = getLanguage();
+
+	let negativePattern = '-{number}';
+	if (language === 'ar-sa') {
```
margaree: it looks like this also applies to just `ar`

dlockhart: Hmm yeah good call. Actually looking through here, I think I should be applying almost all of these number format settings based on the base locale and not the full code. Will fix.

lib/dateTime.js
```diff
@@ -0,0 +1,577 @@
...
+	let weekendStartDay = 6;
+	let weekendEndDay = 0;
+
+	switch (baseLanguage) {
+		case 'ar':
+			dateFormats = ['dddd, d MMMM, yyyy', 'dd MMMM, yyyy', 'dd/MM/yyyy', 'MMMM, yyyy', 'd MMMM'];
+			dayPeriods = ['ص', 'م'];
+			months[0] = months[1] = ['يناير', 'فبراير', 'مارس', 'أبريل', 'مايو', 'يونيو', 'يوليو', 'أغسطس', 'سبتمبر', 'أكتوبر', 'نوفمبر', 'ديسمبر'];
+			days = [
+				['الأحد', 'الإثنين', 'الثلاثاء', 'الأربعاء', 'الخميس', 'الجمعة', 'السبت'],
```
margaree: It looks like the translation for Sunday has been moved to the last position, and Saturday is now first in the array. Is this expected?

lib/dateTime.js
```diff
@@ -0,0 +1,577 @@
...
+	const match = reSeparator.exec(descriptor.formats.dateFormats.short);
+	if (match !== null) {
+		return match[0];
+	}
+	return '/';
+}
+
+function getTimeFormat(hour24, language, baseLanguage) {
+	let timeFormat = hour24 ? 'HH:mm' : 'h:mm tt';
+	if (baseLanguage === 'ja' || baseLanguage === 'pt' || baseLanguage === 'zh') {
```
svanherk: Does `es` also fit under this category?

svanherk: There also seems to be a difference between `pt` and `pt-BR`, unless I'm misunderstanding

lib/number.js
dlockhart: I've fixed this (and a couple other issues) and added tests for all the base languages.

lib/dateTime.js
dlockhart: Looks like this is just a bug in GitHub's web UI -- in the actual source code it's first.

lib/dateTime.js
dlockhart: Right, so I'm actually not sure what to do here. If these changes should match what was previously in the library, you're correct.

The LMS has a DB table with all this stuff in it, but it only has entries for the specific locales -- `es-MX` and `pt-BR`. There's [this library](https://github.com/andyearnshaw/Intl.js/tree/master/locale-data/json) that's used as a polyfill and I think was what we originally used to populate these values, but the LMS isn't even consistent with it.

I think I'll fix this so that we're at least consistent with how things used to be, but in reality I'm not sure how these base languages should work (what regional locale "wins"?), or if we even really care since the LMS should never send down one of these language values anyway.

svanherk: Right yeah, I'm not sure what I should be comparing.  I'm also seeing a difference in https://github.com/Brightspace/intl/pull/31/files#diff-38bec192119c430c4cdc8729f798f951R317 and what used to be there, I don't know what we want to use as a source of truth

dlockhart: Also a good catch -- that one was intentional as the data in here differed from the LMS for pretty much everything for the `en-GB` locale: full date, medium date, short date and month-day.

The other intentional changes were `pt-BR`'s month-year and month-day formats, and `sv-SE`'s month-day. Basically I re-aligned the stuff in here with what was in the LMS's database.

dlockhart: Fixed the `es` and `pt` cases.

package.json
```diff
@@ -1,24 +1,23 @@
 {
-  "name": "d2l-intl",
-  "version": "2.1.0",
-  "description": "D2L internationalization APIs for number, date and time formatting and parsing.",
-  "main": "dist/index.js",
-  "types": "src/index.d.ts",
-  "module": "src/index.js",
+  "name": "@brightspace-ui/intl",
```
dbatiste: Scoped packages are private by default so this will be needed in order to publish:
```
  "publishConfig": {
    "access": "public"
  },
```
Side note: I believe that if the org does not have the private packages feature, the publish will fail unless public access is explicitly specified.

dlockhart: Right -- thanks, fixed!

lib/dateTime.js
```diff
@@ -0,0 +1,578 @@
...
+			days = [
+				['Sonntag', 'Montag', 'Dienstag', 'Mittwoch', 'Donnerstag', 'Freitag', 'Samstag'],
+				['So.', 'Mo.', 'Di.', 'Mi.', 'Do.', 'Fr.', 'Sa.'],
+				['S', 'M', 'D', 'M', 'D', 'F', 'S']
+			];
+			break;
+		case 'en':
+			break;
+		case 'es':
+			dateFormats = ['dddd d\' de \'MMMM\' de \'yyyy', 'd\' de \'MMMM\' de \'yyyy', 'dd/MM/yyyy', 'MMMM yyyy', 'd\' de \'MMMM'];
```
svanherk: Looks like this is in the same boat with `es` to `es-MX` differences (`'d/M/yyyy'` vs `'dd/MM/yyyy'`).  Worth adding the proper `es` value?

lib/dateTime.js
```diff
@@ -0,0 +1,578 @@
...
+
+	options = options || {};
+	options.format = options.format || 'short';
+
+	let format = descriptor.formats.dateFormats[options.format];
+	if (format === undefined) {
+		format = options.format;
+	}
+
+	switch (format) {
```
margaree: This confused me a bit at first, but I think the idea is that these are other dateFormats that are consistent in their format and don't depend on the locale. Would it make sense to move these to https://github.com/Brightspace/intl/blob/58d25f5ce1ef3b62bb71c3756fd18fd11e639155/lib/dateTime.js#L333 so they're all together? I'm fine with it as is, but it would just put them all in one place (though maybe there's something I'm missing and these are way better here)

lib/dateTime.js
svanherk: Assuming French change was also on purpose (previously we had `fr` and `fr-CA` the same, but the database has them different)

dlockhart: French was an interesting one, as we recently added `fr-FR` to the LMS database, which is the closest we're going to get to a base `fr` locale. So I just used its settings for the base `fr`, which yeah was a change since before we were treating `fr-CA` as the base one. Again, I don't think it really matters what the values are for these base ones... the only time they'll be used is if someone uses a custom locale that isn't a match (e.g. `fr-ZZ`).

lib/dateTime.js
```diff
@@ -0,0 +1,578 @@
...
+			];
+			days = [
+				['domingo', 'segunda-feira', 'terça-feira', 'quarta-feira', 'quinta-feira', 'sexta-feira', 'sábado'],
+				['dom', 'seg', 'ter', 'qua', 'qui', 'sex', 'sáb'],
+				['D', 'S', 'T', 'Q', 'Q', 'S', 'S']
+			];
+			break;
+		case 'sv':
+			dateFormats = ['dddd \'den\' d MMMM yyyy', 'd MMMM yyyy', 'yyyy-MM-dd', 'MMMM yyyy', 'dd MMMM'];
+			dayPeriods = ['fm', 'em'];
```
svanherk: These were `FM` and `EM` before.  I haven't been able to track down where this lives in the database, was this a conscious change to match?  We seem to have left `en` capitalized

dlockhart: I was [going off this](https://github.com/andyearnshaw/Intl.js/blob/master/locale-data/json/sv-SE.json#L163) which has them lowercase. Unfortunately there's nothing in the LMS and no way to test it out since trying to format times for locales that use 24-hour clocks breaks (see #32).

lib/dateTime.js
dlockhart: Fixed.

lib/dateTime.js
dlockhart: I like that idea -- will do. Originally yeah it was handled separately since they're not "standard" formats... but now that everything isn't broken out across many files I think just moving it up to the descriptor should be fine.

lib/dateTime.js
```diff
@@ -0,0 +1,578 @@
...
+		case 'en-gb':
+			dateFormats = ['dddd, d MMMM yyyy', 'dd MMMM yyyy', 'dd/MM/yyyy', 'MMMM yyyy', 'd MMMM'];
+			break;
+		case 'fr-ca':
+			dateFormats[1] = 'MMM d yyyy';
+			dateFormats[2] = 'yyyy-MM-dd';
+			dateFormats[4] = 'MMMM d';
+			firstDayOfWeek = 0;
+			break;
+		case 'zh-tw':
```
svanherk: Looks like this might also need a custom `days` array?

lib/number.js
```diff
@@ -0,0 +1,319 @@
...
+
+	let negativePattern = '-{number}';
+	if (baseLanguage === 'ar') {
+		negativePattern = '{number}-';
+	}
+
+	let percentPattern = '{number} %';
+	let percentNegativePattern = '-{number} %';
+	switch (baseLanguage) {
+		case 'es':
```
svanherk: Nit: Another case where `es` and `es-MX` is different

lib/dateTime.js
dlockhart: Correct -- added that in, as well as tests for the missing formats. The reason this didn't get found is that none of the standard date formats use the long day for the ZH locale.

lib/dateTime.js
```diff
@@ -0,0 +1,571 @@
...
+			fullTimeFormat = `${timeFormat} ZZZ`;
+			days[0] = ['星期日', '星期一', '星期二', '星期三', '星期四', '星期五', '星期六'];
```
dlockhart: You'll notice the first entry for Sunday differs from what was in the old JS file. I think it was a bug [based on this](https://github.com/andyearnshaw/Intl.js/blob/master/locale-data/json/zh-Hant-TW.json#L137) and based on the fact that the `日` character is used for Sunday in the other formats.

lib/number.js
dlockhart: Fixed the `es`/`es-MX` case. The Swedish one I'm not uncertain about as the [polyfill](https://github.com/andyearnshaw/Intl.js/blob/master/locale-data/json/sv-SE.json#L282) doesn't reverse the percent symbol at all. I'm going to check with Jeff S... maybe we messed up our LMS data.

svanherk: I think there may also be discrepancies between the decimal and group symbols for those two?

dlockhart: Fixed. I'm going to add these base locale tests back into the v2 branch just to make sure I didn't miss anything else.

svanherk: Sounds good, I am definitely not confident in my quick spot-checking

dlockhart:
I have three outstanding questions for Jeff S, and I'll update this with the answers when I have them:
1. Should the PT short date format have leading zeros for the day and months? The Intl polyfill uses them as does our base PT data, but the LMS does not.
2. Should the SV percent format have the percent at the beginning (LMS) or at the end (Intl polyfill)?
3. Should the SV group separator be a period (LMS) or a space (Intl polyfill)?

PR #33, opened 2020-01-02
new baseline tests for base languages

dlockhart:
This uncovered the following discrepancies in the v3 PR:
- Base `pt` short date format should have leading zeroes (will investigate)
- Base `sv` group separators should be spaces instead of periods (will investigate)
- Base `sv` percent formats aren't inverted (we were aware of this one)


dlockhart:
The old Intl doesn't support `da`, `de` or `nl` -- they're new locales -- so it's just going to fall back to English values for everything. The commented out values are what it _would_ output if it supported them.

dlockhart:
If values are commented out that aren't one of those locales, it's because I'm making a conscious choice to make a change from what it does today to what it'll do in the future (e.g. en-GB).

PR #36, opened 2020-01-07
Add convertUTCToLocalDateTime and convertLocalToUTCDateTime

margaree:
~~`convertUTCToLocalDateTime` takes a UTC date (e.g., `new Date(Date.UTC(2015, ... )`) and returns a `Date` with year/month/day/hour/minute/second corresponding to the `data-timezone` identifier (e.g., `America/Toronto`). This can then be used by `formatDate` etc~~

~~`convertLocalToUTCDateTime` takes the local date and returns what that date is in UTC.
Currently some of the daylight savings tests for this are failing when run with Travis so I'm looking into that.~~

After some discussion it was decided to do the following:
`convertUTCToLocalDateTime` takes an object representing a UTC date/time and returns an object representing a date/time in the local timezone corresponding to the `data-timezone` identifier (e.g., `America/Toronto`)

`convertLocalToUTCDateTime` take an object representing a local date/time and returns an object representing that date/time in UTC based on the local timezone corresponding to the `data-timezone` identifier

The README has these marked as do not use yet since it's possible they will change (e.g., perhaps they will support `Date`, etc)

### DIFF
--- a/lib/dateTime.js
+++ b/lib/dateTime.js
@@ -197,8 +197,10 @@ export function convertLocalToUTCDateTime(date) {
 		return date;
 	}
 	const dateDate = new Date(date.year, date.month - 1, date.date, date.hours, date.minutes, date.seconds);
+	console.log('date here ' + dateDate);
 	const timezone = formatDateString(dateDate).split(' ')[2];
 	const dateStringInTimezone = `${date.month}/${date.date}/${date.year}, ${date.hours}:${date.minutes}:${date.seconds} ${timezone}`;
+	console.log('string ' + dateStringInTimezone);
 	const utcCorrectedDate = new Date(Date.parse(dateStringInTimezone));
 
 	return {
--- a/lib/dateTime.js
+++ b/lib/dateTime.js
@@ -14,7 +14,8 @@ function buildDayPeriodRe(part) {
 	return new RegExp(re, 'i');
 }
 
-function convertDateToTimezone(date, timezone) {
+function formatDateString(date) {
+	const timezone = getDocumentLocaleSettings().timezone.identifier;
 	const options = {
 		year: 'numeric',
 		month: '2-digit',
@@ -22,30 +23,14 @@ function convertDateToTimezone(date, timezone) {
 		hour: '2-digit',
 		minute: '2-digit',
 		second: '2-digit',
-		hour12: false, // use 24-hour time format
+		hour12: false,
 	};
 	if (timezone) {
 		options.timeZone = timezone;
+		options.timeZoneName = 'short';
 	}
-
 	const formatter = new Intl.DateTimeFormat('en-US', options);
-	let formattedDateTime = formatter.format(date);
-
-	if (formattedDateTime.indexOf(',') > -1) {
-		formattedDateTime = formattedDateTime.replace(/,/g, '');
-	}
-
-	const splitDateTime = formattedDateTime.split(' ');
-	const splitDate = splitDateTime[0].split('/'); // 01/31/2019
-	const splitTime = splitDateTime[1].split(':'); // 13:10:30
-
-	return new Date(splitDate[2],
-		splitDate[0] - 1,
-		splitDate[1],
-		splitTime[0],
-		splitTime[1],
-		splitTime[2]
-	);
+	return formatter.format(date);
 }
 
 function getParts() {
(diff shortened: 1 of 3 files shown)


lib/dateTime.js
```diff
@@ -152,6 +186,37 @@ function processPattern(pattern, replacements) {
...
+	let timeChangeAdjustedDate = date;
+	if (timeChangeOffset !== 0) {
+		timeChangeAdjustedDate = new Date(date.getTime() - timeChangeOffset * 60000);
+		dateInTimezone = convertDateToTimezone(timeChangeAdjustedDate, getDocumentLocaleSettings().timezone.identifier);
+	}
+	const dateInUTC = convertDateToTimezone(timeChangeAdjustedDate, 'UTC');
+	const diff = dateInTimezone - dateInUTC;
+
+	// subtract calculated offset from date and create new UTC date based on calculated values
+	const dateWithOffset = new Date(date.getTime() - diff);
```
dlockhart: I'm confused by this too... why isn't just converting it to UTC enough?

lib/dateTime.js
```diff
@@ -152,6 +186,37 @@ function processPattern(pattern, replacements) {
 
 }
 
+export function convertUTCToLocalDateTime(date) {
+	const timezone = getDocumentLocaleSettings().timezone.identifier;
```
dlockhart: Can we make sure that if `timezone.identifier` is blank for whatever reason that the conversion just does nothing?

lib/dateTime.js
```diff
@@ -187,34 +172,42 @@ function processPattern(pattern, replacements) {
...
-	const timezone = getDocumentLocaleSettings().timezone.identifier;
-	return convertDateToTimezone(date, timezone);
+	if (!getDocumentLocaleSettings().timezone.identifier) {
+		return date;
+	}
+	const utcDate = new Date(Date.UTC(date.year, date.month - 1, date.date, date.hours, date.minutes, date.seconds));
+	const formattedDateTime = formatDateString(utcDate);
+	const splitDateTime = formattedDateTime.split(' ');
+	const splitDate = splitDateTime[0].split('/');
+	const splitTime = splitDateTime[1].split(':');
```
dlockhart: Another option here that _might_ be faster would be to use a Regex and bracket groups inside to extract the different parts. This is wrong, but for example something like:

```
const re = /([0-9]{1,2})\/([0-9]{1,2})\/([0-9]{4}) ([0-9]{1,2}):([0-9]{2}):([0-9]{2})/;
const match = formattedDateTime.match(re);
if (match === null || match.length !== 7) {
    // uh... not sure
}
return {
		month: match[1],
		date: match[2],
		year: match[3],
		hours: match[4],
		minutes: match[5],
		seconds: match[6]
	};
```;

File: test/dateTime.js [hunk 4]
```diff
@@ -97,43 +93,103 @@ describe('dateTime', () => {
...
+					date: 25,
+					year: 2015,
+					hours: 12,
+					minutes: 28,
+					seconds: 0
+				};
+
+				documentLocaleSettings.timezone.identifier = 'FAKE';
+				expect(() => {
+					convertLocalToUTCDateTime(UTCDateTime);
```
dlockhart: Is this the browser API that's throwing?

test/dateTime.js
```diff
@@ -15,6 +17,286 @@ describe('dateTime', () => {
...
+		describe('convertUTCToLocalDateTime', () => {
+			it('should return original date if timezone identifier is blank', () => {
+				const UTCDateTime = {
+					month: 8,
+					date: 25,
+					year: 2015,
+					hours: 12,
+					minutes: 28,
+					seconds: 0
+				};
```
dlockhart: Seems like you're reusing the same date over and over again (which is fine), but you could maybe pull it out above and just reference it in your tests to make them smaller?

test/dateTime.js
```diff
@@ -15,6 +17,286 @@ describe('dateTime', () => {
...
+						date: date,
+						year: 2015,
+						hours: hours,
+						minutes: minutes,
+						seconds: 0
+					};
+
+					documentLocaleSettings.timezone.identifier = test.timezone;
+					const result = convertUTCToLocalDateTime(UTCDateTime);
+					const expectedResult = getExpectedResult(date, minutes, hours, test.expectedGMTOffset);
```
dlockhart: Just thinking... if `getExpectedResult()` took in the whole input value and returned a full output value, then the tests could be simplified to:

```
const input = ...;
documentLocaleSettings.timezone.identifier = test.timezone;
const result = convertUTCToLocalDateTime(input);
const expected = getExpectedResult(input, test.expectedGMTOffset);
expect(result).to.deep.equal(expected);
```

test/dateTime.js
margaree: yup

PR #39, opened 2020-01-14
no longer ignoring override of hour24 setting

### LINKED ISSUE #32: Time format strings can't handle arbitrary 12/24 hour clock settings
The time format strings are locale-specific, and are specific to whether that locale uses a 24-hour clock or not.

24-hour time format strings contain `H` or `HH` and no `tt` whereas 12-hour formats contain `h` or `hh` and the `tt` where the AM/PM goes. The problem is that users can choose whether to use a 12 or 24 hour clock independently from their language. So they could choose French (24 hour clock) but then set their clock to 12-hour. That would cause formatted times to be broken.

dlockhart:
Whether or not to use a 24-hour clock can be set via the overrides -- the code was incorrectly not using that value when it was figuring out what format to use for the time format.

I've adjusted the expected values out of the tests, and added commented-out values for things that will change when I fix #32.


lib/dateTime.js
```diff
@@ -228,6 +229,9 @@ export function getDateTimeDescriptor() {
 	if (language === 'zh-tw') {
 		hour24 = false;
 	}
+	if (settings.overrides.date && settings.overrides.date.hour24 !== undefined) {
+		hour24 = settings.overrides.date.hour24;
+	}
```
dlockhart: This was happening down below when it merges everything together, but the code between here and there actually uses the value, so we need to get the correct potentially overridden value here.

test/dateTime.js
```diff
@@ -295,36 +295,36 @@ describe('dateTime', () => {
 		});
 
 		[
```
dlockhart: To make it a bit easier to read, the order that these tests go through in each array is:
early/short/24-hour
early/short/12-hour
early/full/24-hour
early/full/12-hour
late/short/24-hour
late/short/12-hour
late/full/24-hour
late/full/12-hour

test/dateTime.js
```diff
@@ -295,36 +295,36 @@ describe('dateTime', () => {
...
-			{locale: 'de-DE', expect: ['01:28', '01:28', '01:28 EST', '01:28 EST', '13:52', '13:52', '13:52 EST', '13:52 EST']},
-			{locale: 'en', expect: ['1:28 AM', '1:28 AM', '1:28 AM EST', '1:28 AM EST', '1:52 PM', '1:52 PM', '1:52 PM EST', '1:52 PM EST']},
-			{locale: 'en-CA', expect: ['1:28 AM', '1:28 AM', '1:28 AM EST', '1:28 AM EST', '1:52 PM', '1:52 PM', '1:52 PM EST', '1:52 PM EST']},
-			{locale: 'en-GB', expect: ['1:28 AM', '1:28 AM', '1:28 AM EST', '1:28 AM EST', '1:52 PM', '1:52 PM', '1:52 PM EST', '1:52 PM EST']},
-			{locale: 'en-US', expect: ['1:28 AM', '1:28 AM', '1:28 AM EST', '1:28 AM EST', '1:52 PM', '1:52 PM', '1:52 PM EST', '1:52 PM EST']},
-			{locale: 'es', expect: ['1:28', '1:28', '1:28 EST', '1:28 EST', '13:52', '13:52', '13:52 EST', '13:52 EST']},
-			{locale: 'es-MX', expect: ['01:28', '01:28', '01:28 EST', '01:28 EST', '13:52', '13:52', '13:52 EST', '13:52 EST']},
+			{locale: 'ar', expect: ['01:28', '1:28 ص', '01:28 EST', '1:28 ص EST', '13:52', '1:52 م', '13:52 EST', '1:52 م EST']},
+			{locale: 'ar-SA', expect: ['01:28', '1:28 ص', '01:28 EST', '1:28 ص EST', '13:52', '1:52 م', '13:52 EST', '1:52 م EST']},
```
dlockhart: The Arabic is messed up in the GitHub UI diff.

PR #47, opened 2020-02-19
fixing month names

dlockhart:
Using these numeric month names was incorrect. This aligns this library with what the monolith does, which is based on the CLDR database.


lib/dateTime.js
```diff
@@ -335,7 +340,10 @@ export function getDateTimeDescriptor() {
 		case 'nl':
 			dateFormats = ['dddd d MMMM yyyy', 'd MMMM yyyy', 'dd-MM-yyyy', 'MMMM yyyy', 'd MMMM'];
 			dayPeriods = ['a.m.', 'p.m.'];
-			months = [numericMonths, numericMonths];
+			months = [
+				['januari', 'februari', 'maart', 'april', 'mei', 'juni', 'juli', 'augustus', 'september', 'oktober', 'november', 'december'],
+				['jan.', 'feb.', 'mrt.', 'apr.', 'mei', 'jun.', 'jul', 'aug.', 'sep.', 'okt.', 'nov.', 'dec.']
```
margaree: I think `jul` should be `jul.`

dlockhart: Indeed -- thanks!

PR #48, opened 2020-02-27
DA and DE short date format should use dots instead of slashes or hyphens

dlockhart:
I'll be making the same change to the monolith shortly.

References:
- CLDR for da: https://www.unicode.org/cldr/charts/36/summary/da.html#1921
- CLDR for de: https://www.unicode.org/cldr/charts/36/summary/de.html#2003
- Date format by country: https://en.wikipedia.org/wiki/Date_format_by_country


test/dateTime.js
```diff
@@ -618,10 +618,10 @@ describe('dateTime', () => {
 		[
 			{locale: 'ar', expect: ['الثلاثاء, 4 يونيو, 2019', '04 يونيو, 2019', '04/06/2019', 'يونيو, 2019', '4 يونيو', 'الثلاثاء', 'ثلاثاء', 'يونيو', 'يونيو']},
 			{locale: 'ar-SA', expect: ['الثلاثاء, 4 يونيو, 2019', '04 يونيو, 2019', '04/06/2019', 'يونيو, 2019', '4 يونيو', 'الثلاثاء', 'ثلاثاء', 'يونيو', 'يونيو']},
-			{locale: 'da', expect: ['tirsdag den 4. juni 2019', '4. jun.. 2019', '04/06/2019', 'juni 2019', '4. juni', 'tirsdag', 'tir.', 'juni', 'jun.']},
-			{locale: 'da-DK', expect: ['tirsdag den 4. juni 2019', '4. jun.. 2019', '04/06/2019', 'juni 2019', '4. juni', 'tirsdag', 'tir.', 'juni', 'jun.']},
-			{locale: 'de', expect: ['Dienstag 4. Juni 2019', '4. Juni 2019', '04-06-2019', 'Juni 2019', '4. Juni', 'Dienstag', 'Di.', 'Juni', 'Juni']},
-			{locale: 'de-DE', expect: ['Dienstag 4. Juni 2019', '4. Juni 2019', '04-06-2019', 'Juni 2019', '4. Juni', 'Dienstag', 'Di.', 'Juni', 'Juni']},
+			{locale: 'da', expect: ['tirsdag den 4. juni 2019', '4. jun.. 2019', '04.06.2019', 'juni 2019', '4. juni', 'tirsdag', 'tir.', 'juni', 'jun.']},
```
margaree: Not totally related to this PR but I just noticed that `jun..` with the 2 periods. Is this expected? Looks like it's because the period is added in `months` and `dateFormats` for `da`.

dlockhart: I know it's strange eh. It's because the format string is `d. MMM. yyyy` (with periods in it) and then MMM = `jun.` so it ends up with an extra period. I'm not sure how it's actually supposed to work...

PR #68, opened 2020-10-20
Adding shortMonthDay format

ChrisKraljevicD2L:
I hope I've done this correctly.

Lessons uses [`react-intl`'s](https://formatjs.io/docs/react-intl/api/#formatdate) `{ month: 'short', day: 'numeric' }` date format for displaying pacing dates. We want to keep supporting that format as we're switching over to using `BrightspaceUI/intl` and displaying more dates on content.

If I need to reach out to someone for correct `shortMonthDay` formatting, please let me know.


lib/dateTime.js
```diff
@@ -537,20 +537,20 @@ export function getDateTimeDescriptor() {
...
 		case 'ja':
-			dateFormats = ['yyyy年M月d日', 'yyyy年M月d日', 'yyyy/MM/dd', 'yyyy年M月', 'M月d日'];
+			dateFormats = ['yyyy年M月d日', 'yyyy年M月d日', 'yyyy/MM/dd', 'yyyy年M月', 'M月d日', 'M月d日'];
 			dayPeriods = ['午前', '午後'];
 			months[0] = months[1] = ['1 月', '2 月', '3 月', '4 月', '5 月', '6 月', '7 月', '8 月', '9 月', '10 月', '11 月', '12 月'];
 			days[0] = days[1] = days[2] = ['日', '月', '火', '水', '木', '金', '土'];
 			break;
 		case 'ko':
-			dateFormats = ['yyyy년 M월 d일 dddd', 'yyyy년 M월 d일', 'yyyy-MM-dd', 'yyyy년 M월', 'M월 d일'];
+			dateFormats = ['yyyy년 M월 d일 dddd', 'yyyy년 M월 d일', 'yyyy-MM-dd', 'yyyy년 M월', 'M월 d일', 'M월 d일'];
```
dlockhart, line 546: The CLDR says this should be: `MMM d일`

lib/dateTime.js
```diff
@@ -620,6 +620,7 @@ export function getDateTimeDescriptor() {
 			dateFormats[1] = 'MMM d yyyy';
 			dateFormats[2] = 'yyyy-MM-dd';
 			dateFormats[4] = 'MMMM d';
+			dateFormats[5] = 'MMM d';
```
dlockhart, line 623: The CLDR doesn't have a special entry here for `fr-CA`, but I'm good either way.

dlockhart, line 623: The CLDR doesn't have a special entry here for `fr-CA`, but I'm good either way.

lib/dateTime.js
ChrisKraljevicD2L: Odd, tests still passed after I changed this without changing the expected date format 🤔

PR #69, opened 2020-11-16
Switching from Travis CI to GitHub Actions

dlockhart:
Deploy part still waiting for the semantic-release action to be merged.


.github/workflow/release.yml
```diff
@@ -0,0 +1,24 @@
+name: Release
+on:
+  push:
+    branches:
+      - master
```
svanherk, line 5: Should we add the regex for version branches here since this is using `semantic-release`?

README.md
```diff
@@ -3,7 +3,6 @@
 [![NPM version](https://img.shields.io/npm/v/@brightspace-ui/intl.svg)](https://www.npmjs.org/package/@brightspace-ui/intl)
 [![NPM downloads](https://img.shields.io/npm/dt/@brightspace-ui/intl.svg)](https://www.npmjs.com/package/@brightspace-ui/intl)
 [![Dependabot badge](https://flat.badgen.net/dependabot/BrightspaceUI/intl?icon=dependabot)](https://app.dependabot.com/)
-[![Build status](https://travis-ci.com/BrightspaceUI/intl.svg?branch=master)](https://travis-ci.com/BrightspaceUI/intl)
```
svanherk, line 6: Minor: We can add a Github Action status badge if we want: https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/adding-a-workflow-status-badge

I'm indifferent - the status was never the useful part for me, but the link to Travis was.  Since now we don't need to leave the repo anymore... 🤷

dlockhart: Good call. I removed the Dependabot badge as it's actually broken and GitHub knows it.

svanherk (approved):
This looks like what I was imagining after our meeting!

PR #71, opened 2020-11-17
fixing badge [skip ci]

dlockhart:


svanherk (approved):
Huh... would not expect case to matter

PR #74, opened 2020-11-24
Bug fix: support more timezones

margaree:
Timezones source of truth is now this list https://www.timeanddate.com/time/zones/. I added and corrected where needed and added support for duplicates, which was where we had problems with Halifax.

### DIFF
--- a/lib/dateTime.js
+++ b/lib/dateTime.js
@@ -3,12 +3,11 @@ import { getDocumentLocaleSettings, getLanguage, merge } from './common.js';
 const hour24locales = ['da', 'de', 'es', 'fr', 'nl', 'pt', 'sv', 'tr', 'zh'];
 const mondayFirstDayLocales = ['da', 'de', 'fr', 'nl', 'sv', 'tr'];
 
-// leaving out the following since do not throw and have multiple possible offsets: AMT, AST, BST, CDT, CST, ECT, GST, IST, LHST, MST, PST, SST
 const timezoneOffsetMap = {
 	'ACDT': '+1030',
 	'ACST': '+0930',
+	'ACT': '-0500',
 	'ACWST': '+0845',
-	'ADT': '-0300',
 	'AEDT': '+1100',
 	'AEST': '+1000',
 	'AET': '+1000',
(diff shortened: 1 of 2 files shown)


lib/dateTime.js
```diff
@@ -216,16 +308,82 @@ function formatDateString(date) {
...
+			return longTimezone === 'America/Havana' ? '-0400' : '-0500';
+		case 'CST':
+			// Cuba Standard Time -0500 (America/Havana)
+			// Central Standard Time -0600 (America)
+			// China Standard Time +0800 (Asia)
+			if (longTimezone === 'America/Havana') return '-0500';
+			else if (longTimezone.includes('America')) return '-0600';
+			else return '+0800';
+		case 'GST':
+			// South Georgia and the South Sandwich Islands Time -0200 (Atlantic)
```
dlockhart, line 360: I'm totally going to the South Sandwich Islands when this pandemic is over. 🇬🇸

lib/dateTime.js
```diff
@@ -3,73 +3,96 @@ import { getDocumentLocaleSettings, getLanguage, merge } from './common.js';
 const hour24locales = ['da', 'de', 'es', 'fr', 'nl', 'pt', 'sv', 'tr', 'zh'];
 const mondayFirstDayLocales = ['da', 'de', 'fr', 'nl', 'sv', 'tr'];
 
-// leaving out the following since do not throw and have multiple possible offsets: AMT, AST, BST, CDT, CST, ECT, GST, IST, LHST, MST, PST, SST
```
dlockhart, line 6: Could you include that URL where you got this data in a comment? Just in case we need to go back in here later.

lib/dateTime.js
```diff
@@ -216,16 +308,82 @@ function formatDateString(date) {
...
+		case 'PYT':
+			// Paraguary Time -0400 (South America)
+			// Pyongyan Time +0830 (Asia)
+			return longTimezone.includes('Asia') ? '+0830' : '-0400';
+		case 'WST':
+			// West Samoa Time +1400 (Pacific)
+			// Western Sahara Summer Time +0100 (Africa)
+			return longTimezone.includes('Pacific') ? '+1400' : '+0100';
+		default:
+			throw new Error(`Invalid timezone: unable to retrieve timezone offset for ${longTimezone}`);
```
dlockhart, line 383: Are we reasonably confident that this isn't going to throw very often? Like I guess previously this would have resulted in an `undefined` timezone which might have blown up elsewhere in the code.

lib/dateTime.js
```diff
@@ -216,16 +308,82 @@ function formatDateString(date) {
...
 	}
 	return dateString;
 }
 
+function getTimezoneFromDuplicatedAbbreviation(abbrTimezone) {
+	const longTimezone = getDocumentLocaleSettings().timezone.identifier;
+	switch (abbrTimezone) {
+		case 'ADT':
+			// Arabia Daylight Time +0400 (Asia)
+			// Atlantic Daylight Time -0300 (America)
```
dlockhart, line 327: In the web site you listed, it shows other abbreviations for Halifax -- ADST, AST and HAA -- I'm assuming those don't get used by the browsers as they're not in the map?

dlockhart (approved):
It's so crazy that there's ambiguity in the abbreviation... like they couldn't just come up with unique abbreviations?! Seems like an opportunity to standardize.

lib/dateTime.js
margaree: It shouldn't blow up anymore than before. Before if the timezone was `undefined` then it would fail in `getDateStringWithTimezone` slightly later in the same function so I figured adding this would help make debugging more clear. I think throwing somewhere in here is probably what we want since we wouldn't want `convertLocalToUTCDateTime` to return an incorrect value.

lib/dateTime.js
margaree: I think those aren't used very frequently, though I can add them just in case. I did start adding the alternatives but it must have been later on in the file and I missed going back to the top.

lib/dateTime.js
dlockhart: Sounds good. Yeah at least now we could actively search for this error in the wild!

dlockhart (approved):

PR #76, opened 2020-12-03
feat: add formatTimestamp method

johngwilkinson:
Testing:
- unit tests
- npm link and used new method via BSI in our web component
- tested several timezone settings in Account Settings

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ const value = parseNumber('-8 942,39'); // -> -8942.39 in fr-CA
 
 Dates and times can be formatted in the user's locale using `formatDate`, `formatTime` and `formatDateTime`.
 
+Timestamps (milliseconds since the epoch) can be formatted in the user's locale and timezone using `formatTimestamp`.
+
 Combined dates and times are formatted using `formatDateTime`:
 
 ```javascript
@@ -94,6 +96,18 @@ Options:
   - **full**: long weekday, month names and timezone. e.g. `'Wednesday, September 23, 2015 1:25 PM EST'`
   - **medium**: long month names. e.g. `'September 23, 2015 1:25 PM'`
   - **short**: abbreviated date format. e.g. `'9/23/2015 1:25 PM'`
+
+To format a **timestamp** as a date and time:
+
+```javascript
+const dateString = formatTimestamp(
+	1607097863123,
+	[options]
+);
+```
+
+Options are the same as for formatDateTime: this method is converts the timestamp to a `Date` in the user's
+configured time zone, then returns the results of passing this date to `formatDateTime`.
 
 To format a **date only** (without the time portion), use `formatDate`:
 
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ const dateString = formatTimestamp(
 );
 ```
 
-Options are the same as for formatDateTime: this method is converts the timestamp to a `Date` in the user's
+Options are the same as for `formatDateTime`; this method converts the timestamp to a `Date` in the user's
 configured time zone, then returns the results of passing this date to `formatDateTime`.
 
 To format a **date only** (without the time portion), use `formatDate`:
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ const value = parseNumber('-8 942,39'); // -> -8942.39 in fr-CA
 
 Dates and times can be formatted in the user's locale using `formatDate`, `formatTime` and `formatDateTime`.
 
-Timestamps (milliseconds since the epoch) can be formatted in the user's locale and timezone using `formatTimestamp`.
+Timestamps (milliseconds since the epoch) can be formatted in the user's locale and timezone using `formatDateTimeFromTimestamp`.
 
 Combined dates and times are formatted using `formatDateTime`:
 
@@ -100,7 +100,7 @@ Options:
 To format a **timestamp** as a date and time:
 
 ```javascript
-const dateString = formatTimestamp(
+const dateString = formatDateTimeFromTimestamp(
 	1607097863123,
 	[options]
 );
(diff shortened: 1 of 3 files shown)


margaree:
Thanks for the contribution! It seems like it would be more in line with the current structure of DateTime to instead create a convert method (e.g., `convertUTCTimestampToLocalDateTime`) and then use the result of that in `formatDateTime` since that seems to be the end goal anyways. For example, your usage could become something like:
```
const localDateTime = convertUTCTimestampToLocalDateTime(timestamp);
const result = formatDateTime(new Date(localDateTime.year, localDateTime.month - 1, localDateTime.date, localDateTime.hours, localDateTime.minutes, localDateTime.seconds));
```

I also just want to confirm your timestamp use case - when I had thought timestamp I had thought you'd be sending in something like `"1437827280"` but based on your tests it looks like you'll just be sending in a timestamp using `Date.UTC`. Either way can you document this in the README?

johngwilkinson:
> Thanks for the contribution! It seems like it would be more in line with the current structure of DateTime to instead create a convert method (e.g., `convertUTCTimestampToLocalDateTime`) and then use the result of that in `formatDateTime` since that seems to be the end goal anyways. For example, your usage could become something like:
> 
> ```
> const localDateTime = convertUTCTimestampToLocalDateTime(timestamp);
> const result = formatDateTime(new Date(localDateTime.year, localDateTime.month - 1, localDateTime.date, localDateTime.hours, localDateTime.minutes, localDateTime.seconds));
> ```
> 
> I also just want to confirm your timestamp use case - when I had thought timestamp I had thought you'd be sending in something like `"1437827280"` but based on your tests it looks like you'll just be sending in a timestamp using `Date.UTC`. Either way can you document this in the README?

I should update the README, yes. `Date.UTC` produces a timestamp like `"1437827280"` (probably more digits), so your first thought about the use case was right.

I'm not sure why an interface that makes the consumer write all that code would be preferable. I'd suggest we should be trying to capture the one way that timestamps should always be rendered, and that's what this method attempts to do: take a timestamp, return the correct display string. That way there's no question about the right way to do it: the lib captures the solved problem. Not sure consistency with the current structure is an argument for making it more difficult to use.

Would a different name for the function make the intention clearer maybe?

lib/dateTime.js
```diff
@@ -622,16 +622,12 @@ function processPattern(pattern, replacements) {
 
 }
 
-export function convertUTCToLocalDateTime(date) {
-	if (!getDocumentLocaleSettings().timezone.identifier) {
-		return date;
-	}
-	const utcDate = new Date(Date.UTC(date.year, date.month - 1, date.date, date.hours, date.minutes, date.seconds));
+function convertJsDateToLocalDateTime(utcDate) {
```
margaree, line 625: This should be in alphabetical order with the other helper functions. Is the change to returning either false or date object in order to avoid an extra `new Date` below here https://github.com/BrightspaceUI/intl/pull/76/files#diff-bfa90af68e38b31e2c7389bc93445f3bb17696889e6665f50c90c27d6813903cR1124 ? It seems like we should avoid a function returning two possible types since if this is used in the future a user would have to know to look out for that.

README.md
```diff
@@ -95,6 +97,18 @@ Options:
...
+To format a **timestamp** as a date and time:
+
+```javascript
+const dateString = formatTimestamp(
+	1607097863123,
+	[options]
+);
+```
+
+Options are the same as for formatDateTime: this method is converts the timestamp to a `Date` in the user's
```
margaree, line 109:
```suggestion
Options are the same as for `formatDateTime`; this method converts the timestamp to a `Date` in the user's
```

margaree (commented):
Thanks for the clarification around timestamp input.

I see your point about having the consumer write less code but I would like a change in the function name, since the `format*` functions in `dateTime` all take a js Date and return whatever the latter part of the function name is (e.g., `formatTime` returns a Date formatted as a time etc) so to me the current name would imply formatting a Date as a timestamp. My suggestion is `formatDateTimeFromTimestamp`.

lib/dateTime.js
johngwilkinson: Alphabetical: ✔️ 

For consistent return, here are some possibilities:
1. the function returns `null` instead of `false` - this is the smallest change
2. the function also takes a default value, and returns it if it can't do the conversion - in this case, `formatDateTimeFromTimestamp` ends up building a default object from a javascript `Date` only to immediately convert it back to the same `Date` - which is a bit on the strange and verbose side, but okay I guess
3. the function returns an object `{ result, isSuccess }`, and callers should use the result only if `isSuccess` is true - this is like the `Try...` pattern in C#, but in javascript I believe it's more idiomatic to do (1)
4. the function throws if it can't convert - seems like overkill

Do you have a preference among those?

margaree: I think 1 or 2 would make the most sense. I'm fine with 1.

johngwilkinson: Okay, done, and function renamed.

test/dateTime.js
```diff
@@ -5,6 +5,7 @@ import {
 	formatDate,
 	formatDateTime,
 	formatTime,
+	formatDateTimeFromTimestamp,
```
margaree, line 8: nit: alphabetization

test/dateTime.js
```diff
@@ -834,4 +835,46 @@ describe('dateTime', () => {
 
 	});
 
+	describe('formatTimestamp', () => {
```
margaree, line 838: nit: should be `formatDateTimeFromTimestamp`

margaree (approved):
LGTM. There are a couple of nits but if you really want to get this merged then I have some clean up tweaks I want to make to other things so I can fix those up in there.

johngwilkinson:
> LGTM. There are a couple of nits but if you really want to get this merged then I have some clean up tweaks I want to make to other things so I can fix those up in there.

That's a kind offer, thanks!

PR #83, opened 2021-03-17
fix: switching to use native formatNumber API

dlockhart:
We're still seeing floating point rounding problems when the number of digits + the requested precision (15 digits from the LMS) exceeds the maximum integer size in JavaScript (`2^53`). For example, `99,999` with a precision of 15 digits resulted in `99,000 x 10^15` being greater than `2^53` so rounding occurs and precision is lost.

So I decided to go back to the drawing board and do something with our Intl stuff that I've wanted to do for a while -- just call into the native `Intl.MessageFormat` APIs! What a concept! The trick here is that we just pass in a known locale (the `en-US` one) where we know the decimal separator will be a period. Then we can grab the stuff before/after the period and re-format it using our crazy rules.

Bonus: it supports max/min fraction digits and handles all the rounding and zero-padding for us!

[Browser support](https://caniuse.com/mdn-javascript_builtins_intl_numberformat_format) seems good: IE11, Edge 12+, Firefox 29+, Chrome 24+ Safari 10+.

### DIFF
--- a/lib/number.js
+++ b/lib/number.js
@@ -43,18 +43,6 @@ function formatPositiveInteger(value, descriptor) {
 	return ret;
 }
 
-function formatExponentialDecimal(value) {
-	// Get a value should be like "1.25e-8"
-	const pieces = value.split('e-');
-	let ret = pieces[0].replace('.', '');
-	const zeroCount = parseInt(pieces[1]) - 1;
-
-	for (let i = 0; i < zeroCount; i++) {
-		ret = '0' + ret;
-	}
-	return ret;
-}
-
 function validateFormatOptions(options) {
 
 	options = options || {};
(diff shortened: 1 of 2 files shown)


lib/number.js
```diff
@@ -192,41 +180,23 @@ function formatDecimal(value, options) {
...
-		decimalStr = '';
-	}
+	).format(value);
 
-	if (decimalStr !== null) {
-		while (decimalStr.length < options.minimumFractionDigits) {
-			decimalStr += '0';
-		}
-		ret += descriptor.symbols.decimal + decimalStr;
+	let ret = formatPositiveInteger(parseInt(strValue), descriptor);
```
dlockhart, line 195: If we didn't need to support our crazy grouping formats (the groups can be different sizes) this could be simplified further.

dbatiste: I guess the same goes for the negative eh?

dbatiste (approved):

lib/number.js
dlockhart: Yes, although swapping in the negative symbol is pretty easy... but yeah I'd love to just rely on the native API even for that. Maybe one day we can do an analysis of what the native API produces and see if we can leverage it... the trick is you need to pass it a single locale that it'll use for the group, decimal AND negative symbol, and we let users customize all 3 of those things separately. But maybe for each of those 9 combinations there's a single locale we could map them to, who knows!

PR #85, opened 2021-03-30
feat: adding support to fr-ON custom locale

dlockhart:
Was hoping we'd avoid needing to deal with the non-standard `fr-ON` locale in here, but because we (correctly) inherit date/time/number settings from the base `fr` instead of `fr-CA`, we need to handle it specially.


bearfriend:
DO we actually just want to change `fr` to be a copy of `fr-CA`, since that's what our other language/locale fallbacks do? That way we'd avoid `fr-ON` altogether and `fr-*` would get both `fr-CA` translations and dates etc.

bearfriend (approved):
Ugh. This does make sense, I'm just now wondering why `fr-CA` should even be the fallback for translations.

dlockhart:
> Do we actually just want to change fr to be a copy of fr-CA

That would certainly make the change easier, and also be consistent with how the monolith handles things. But generally I don't think it's how these things should work... but who knows. The monolith works this way because for many years, we only had clients in `fr-CA` -- it's only recently that we've needed to support `fr-FR` (same for `es-MX` and `es-ES`).

PR #90, opened 2021-07-14
Add default branch to semantic release, update docs

GeorgePlukov:


.github/workflows/release.yml
```diff
@@ -21,6 +21,7 @@ jobs:
       - name: Semantic Release
         uses: BrightspaceUI/actions/semantic-release@master
```
dlockhart, line 22: Eek... we'll have to be careful here when we rename the `actions` repo. Although if it creates an alias maybe this is OK?

Once the policy was approved and this work was done, I was going to go into the actions repo anyway and inverse the default, so that to use "main" you wouldn't need to pass anything. But that'll involve going through every repo and passing in `master` first.

svanherk: > Although if it creates an alias maybe this is OK?

Yeah we should test that.  If not, we could just keep the `master` branch alongside `main` and keep them in sync until we have everyone using it moved over

dlockhart: Yeah, and seeing as my plan is to go through them all and add `DEFAULT_BRANCH: master` so that I can make "main" default anyway, I could fix that at the same time.

GeorgePlukov: Yeah I was going to mention I wanted to test this. I know that links are supposed to get forwarded to the renamed branch but I'm not sure exactly how it will work in this case. 

In the renaming section github published (https://github.com/github/renaming) there are a couple lines explaining what should happen: 

- `Redirect web requests for the old branch name to the new branch name`
- `Return a "Moved Permanently" response in API requests for the old branch name`

dlockhart (approved):

PR #92, opened 2021-08-09
fix: correct readME about medium format

KevinChenXue:
I found that the 'medium' format actually maps to 'MMM d, yyyy' instead of 'MMMM d, yyyy' for English (and a few other languages). Just fixing the docs to match the implementation


dlockhart (approved):
Thanks!

PR #95, opened 2021-11-26
feat: call listeners after other important attributes change

margaree:


lib/common.js
```diff
@@ -121,6 +121,7 @@ class DocumentLocaleSettings {
 			}
 		}
 		this._overrides = val;
+		this._listeners.forEach((cb) => cb());
```
margaree, line 124: So I wasn't sure of a completely ideal way to do all this. Since `overrides`, `language`, and `fallbackLanguage` can technically get changed through `.fallbackLanguage` etc., they should call the listeners, so I didn't want to duplicate that in `_handleObserverChange`, but it feels a bit weird to have two different ways that the calls are happening.

test/common.js
```diff
@@ -124,61 +124,98 @@ describe('common', () => {
 
 	describe('document mutations', () => {
 		it('should update language if "lang" gets set', (done) => {
-			documentLocaleSettings.addChangeListener(() => {
+			const cb = () => {
```
margaree, line 127: Since `afterEach` was resetting the attributes, these listeners were getting called again, causing an error about `done` being called multiple times.

dlockhart (approved):
This looks good to me! Only question in my mind is whether we should be wrapping `timezone` and `oslo` in setters like the others and calling the listeners from there, but both approaches work.

PR #96, opened 2021-11-26
Added helpers for formatting timezones

ronan-f:
I would like to propose we add the following methods to the date time helpers:

1. `formatDateFromTimestamp`
2. `formatTimeFromTimestamp`

This seems consistent with the other formatting helpers which provide options for formatting the date or time individually and allows more flexibility for consumers using timestamps.

Another thought I had was to expose the `parseLocalDateTimeFromTimestamp` method and allow consumers to format the resulting `Date` object however they want. I'm not sure if that would be a preferable option in this case.

### DIFF
--- a/lib/dateTime.js
+++ b/lib/dateTime.js
@@ -1128,24 +1128,24 @@ export function formatDateTime(date, options) {
 
 }
 
-function parseLocalDateTimeFromTimestamp(timestamp, options) {
+function parseLocalDateTimeFromTimestamp(timestamp) {
 	const utcDate = new Date(timestamp);
 	const local = convertJsDateToLocalDateTime(utcDate);
-	if (!local) return formatDateTime(utcDate, options);
+	if (!local) return utcDate;
 	return new Date(local.year, local.month - 1, local.date, local.hours, local.minutes, local.seconds);
 }
 
 export function formatDateTimeFromTimestamp(timestamp, options) {
-	const localDate = parseLocalDateTimeFromTimestamp(timestamp, options);
-	return formatDateTime(localDate, options);
+	const date = parseLocalDateTimeFromTimestamp(timestamp);
+	return formatDateTime(date, options);
 }
 
 export function formatDateFromTimestamp(timestamp, options) {
-	const localDate = parseLocalDateTimeFromTimestamp(timestamp, options);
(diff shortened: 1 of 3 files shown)


lib/dateTime.js
```diff
@@ -1128,10 +1128,24 @@ export function formatDateTime(date, options) {
 
 }
 
-export function formatDateTimeFromTimestamp(timestamp, options) {
+function parseLocalDateTimeFromTimestamp(timestamp, options) {
```
ronan-f, line 1131: Methods which returned JS `Date` objects seemed to be named `parse<dateFormat>` so I tried to keep that consistent here.

lib/dateTime.js
```diff
@@ -1128,10 +1128,24 @@ export function formatDateTime(date, options) {
...
 	const utcDate = new Date(timestamp);
 	const local = convertJsDateToLocalDateTime(utcDate);
 	if (!local) return formatDateTime(utcDate, options);
```
margaree, line 1134: I think you'd want this to return just `utcDate` in this case, since `formatDateTime` is then applied to whatever is returned.

lib/dateTime.js
```diff
@@ -1128,10 +1128,24 @@ export function formatDateTime(date, options) {
...
-	const localDate = new Date(local.year, local.month - 1, local.date, local.hours, local.minutes, local.seconds);
+	return new Date(local.year, local.month - 1, local.date, local.hours, local.minutes, local.seconds);
+}
+
+export function formatDateTimeFromTimestamp(timestamp, options) {
+	const localDate = parseLocalDateTimeFromTimestamp(timestamp, options);
```
margaree, line 1139: Since this could return utc or local date, would probably just want to call this `const date`

lib/dateTime.js
ronan-f: Ah. Good point. I've made that change now.

lib/dateTime.js
ronan-f: Updated

margaree (approved):
Looks good! Do you have merging permissions? If so, when you merge if you append `feat:` to the merge commit message then it will update the version (https://github.com/BrightspaceUI/intl#versioning--releasing)

PR #97, opened 2021-12-14
DE46356 - Update dateTime.js in Brightspace UI with updated Canadian/Ontarian French date formats

joshhoey:
- Removing "le" from long date format of fr-ca and fr-fr ("le" remains in fr-on).
- Making other elements consistent with formats in `LOCALE_CULTURE`.

### DIFF
--- a/lib/dateTime.js
+++ b/lib/dateTime.js
@@ -884,7 +884,7 @@ export function getDateTimeDescriptor() {
 			firstDayOfWeek = 0;
 			break;
 		case 'fr-on':
-			dateFormats[0] = 'dddd\' le \'d MMMM yyyy'; 
+			dateFormats[0] = 'dddd\' le \'d MMMM yyyy';
 			dateFormats[1] = 'MMM d yyyy';
 			dateFormats[2] = 'yyyy-MM-dd';
 			firstDayOfWeek = 0;
--- a/lib/dateTime.js
+++ b/lib/dateTime.js
@@ -788,7 +788,7 @@ export function getDateTimeDescriptor() {
 			];
 			break;
 		case 'fr':
-			dateFormats = ['dddd\' le \'d MMMM yyyy', 'd MMM yyyy', 'dd/MM/yyyy', 'MMMM yyyy', 'd MMMM', 'd MMM'];
+			dateFormats = ['dddd d MMMM yyyy', 'd MMM yyyy', 'dd/MM/yyyy', 'MMMM yyyy', 'd MMMM', 'd MMM'];
 			months = [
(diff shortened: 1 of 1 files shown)


davewagler-d2l (approved):

lib/dateTime.js
```diff
@@ -877,13 +877,18 @@ export function getDateTimeDescriptor() {
...
 			dateFormats[4] = 'MMMM d';
 			dateFormats[5] = 'MMM d';
 			firstDayOfWeek = 0;
 			break;
+		case 'fr-on':
+			dateFormats[0] = 'dddd\' le \'d MMMM yyyy'; 
+			dateFormats[1] = 'MMM d yyyy';
+			dateFormats[2] = 'yyyy-MM-dd';
+			firstDayOfWeek = 0;
+			break;
```
dlockhart, line 891: Not sure why I had to press a button to allow the CI to run... but it caught a linting error in here and I'm guessing the unit tests will fail due to the "le" change.

joshhoey: Thanks, I just removed the trailing space. We'll see how this goes. 👍

dlockhart: Oh I know why it needs me to approve your CI runs -- it's because this is a fork. If you're a member of the BrightspaceUI organization you should have Write access to the repo.

dlockhart: As expected, some of the unit tests are failing. There are a couple that I wouldn't have expected (`4 juin` vs. `juin 4`) so probably something to have a look at there.

joshhoey: I needed to change some tests in 'lms', too.

I have no experience with the BrightspaceUI code (I'm not a member!) so I'll take a look around for them.

lib/dateTime.js
```diff
@@ -877,13 +877,18 @@ export function getDateTimeDescriptor() {
...
 			dateFormats[1] = 'MMM d yyyy';
 			dateFormats[2] = 'yyyy-MM-dd';
 			dateFormats[4] = 'MMMM d';
 			dateFormats[5] = 'MMM d';
 			firstDayOfWeek = 0;
 			break;
+		case 'fr-on':
+			dateFormats[0] = 'dddd\' le \'d MMMM yyyy';
+			dateFormats[1] = 'MMM d yyyy';
+			dateFormats[2] = 'yyyy-MM-dd';
```
dlockhart, line 889: This is also going to change `dateFormats[4]` from `MMMM d` to `d MMMM` and `dateFormats[5]` from `MMM d` to `d MMM` -- did you intend to make that change as well?

joshhoey:
Required changes to two files, so I'm closing this PR in favour of the following:
- #98

PR #98, opened 2021-12-15
DE46356 - Update BrightspaceUI to match core French date changes

joshhoey:
- Removing "le" from long date format of fr-ca and fr-fr ("le" remains in fr-on).
- Making other elements consistent with formats in `LOCALE_CULTURE`.
- Updating tests to match new values.


dlockhart (approved):
LGTM.

PR #102, opened 2022-03-01
feat: Add support for Hindi

margaree:
This is entirely based on what's in the LMS.

Differences between lms and https://www.unicode.org/cldr/cldr-aux/charts/25/summary/hi.html
- Apr: lms: `अप्रैल`, cldr: `अप्रै`
- Jul: lms: `जुलाई `, cldr: `जुला`
- Thursday: lms: `गुरूवार`, cldr: `गुरुवार`
- `['S', 'M', 'T', 'W', 'T', 'F', 'S']` lms not translated, cldr: `['र', 'सो', 'मं', 'बु', 'गु', 'शु', 'श']`

The `dateFormats` from the lms are identical to `es` and based on the date formats [here](https://www.unicode.org/cldr/cldr-aux/charts/25/summary/hi.html#1482) they might be incorrect ~(e.g., looks like Hindi uses m/d/y, not d/m/y)~ (this was incorrect, but the formats are still quite different and don't contain de)


dlockhart (approved):
This LGTM, assuming you have all the info you need from the localization folks.

PR #107, opened 2022-07-05
feat: Add list getSeparator utility

eKoopmans:
This PR adds a utility for getting the correct list separator (e.g. ", ") for the current locale. This is required for the new `object-property-list` component:

> To get the object-property-list component to read correctly on screen-readers, we render "hidden commas" between items. For this to work correctly across locales, we need a way to determine the appropriate list separator for the current locale.

Rally: https://rally1.rallydev.com/#/?detail=/userstory/641658870535&fdp=true

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,20 @@ import {formatFileSize} from '@brightspace-ui/intl/lib/fileSize.js';
 const fileSize = formatFileSize(100); // -> '100 bytes' in en-US
 ```
 
+## List Formatting
+
+Use `getSeparator` to get the appropriate list separator for the current locale. This is a separator that would be used in spoken language; note that the separator includes a space, for locales where it is appropriate.
+
+```javascript
+import {getSeparator} from '@brightspace-ui/intl/lib/list.js';
+
+const separator = getSeparator(); // -> ', ' in en-US
+const separator = getSeparator({ nonBreaking: true }); // -> ',\xa0' in en-US
+```
+
+Options:
+- **nonBreaking**: a Boolean flag, whether to use non-breaking spaces instead of standard spaces; default is `false`
+
 ## Contributing
 
 Contributions are welcome, please submit a pull request!
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@ npm install @brightspace-ui/intl
 Then `import` only the functionality you need:
 
 ```javascript
-import {formatDate, formatTime} from '@brightspace-ui/intl/lib/dateTime.js';
-import {formatNumber, formatPercent} from '@brightspace-ui/intl/lib/number.js';
+import { formatDate, formatTime } from '@brightspace-ui/intl/lib/dateTime.js';
+import { formatNumber, formatPercent } from '@brightspace-ui/intl/lib/number.js';
 ```
 
 All of the APIs will automatically detect the document's language via the `lang` attribute on the `<html>` element. They'll also look for various `data-` attributes that will be present in Brightspace pages to access override and timezone information.
@@ -33,7 +33,7 @@ All of the APIs will automatically detect the document's language via the `lang`
 Integer and decimal numbers can be formatted in the user's locale using `formatNumber`. Percentages can be formatted using `formatPercent`. Use the optional `options` parameter for rounding.
 
 ```javascript
-import {formatNumber, formatPercent} from '@brightspace-ui/intl/lib/number.js';
+import { formatNumber, formatPercent } from '@brightspace-ui/intl/lib/number.js';
 
 const number = formatNumber(8902.72, [options]); // -> '8,902.72' in en-US
 const percent = formatPercent(0.333, [options]); // -> '33.3 %' in en-US
@@ -47,7 +47,7 @@ Options:
 Formatting as an integer (rounded to 0 decimal places):
 
 ```javascript
-import {formatNumber} from '@brightspace-ui/intl/lib/number.js';
+import { formatNumber } from '@brightspace-ui/intl/lib/number.js';
 
 const value = formatNumber(89.72, {
 	maximumFractionDigits: 0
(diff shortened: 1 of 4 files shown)


lib/list.js
```diff
@@ -0,0 +1,21 @@
+import { getLanguage } from './common';
```
dcostiuc, line 1: Looks like you're missing a `.js` here (according due to the check failure).

```suggestion
import { getLanguage } from './common.js';
```

eKoopmans: Yep on it - I'm also adding some unit tests.

lib/list.js
```diff
@@ -0,0 +1,21 @@
...
+	const langTag = getLanguage();
+	const space = nonBreaking ? nbsp : ' ';
+
+	// In this function, most languages return the separator used by the Intl.ListFormat() API with options:
+	//		{ style: 'short', type: 'conjunction' }
+	// 		https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat
+	// Arabic, due to some inconsistencies, returns the separator used by Google Translate.
+
+	if (langTag === 'ar') return `${space}،${space}`;
+	if (['ja', 'zh-cn', 'zh-tw'].includes(langTag)) return '、';
```
dcostiuc, line 19: I was wondering why we didn't put a `${space}` here but I realized that there's already a space in the character `、`, for anyone else wondering.

eKoopmans: Yeah, sort of - that `、` is one single wide character, with its own built-in space. Doesn't seem to use an added space with it.

dcostiuc: Yeah that's what I meant haha. Visually it looks as though a separate space was put after the comma, but it's actually just whitespace embedded within the character.

README.md
```diff
@@ -244,6 +244,20 @@ import {formatFileSize} from '@brightspace-ui/intl/lib/fileSize.js';
 const fileSize = formatFileSize(100); // -> '100 bytes' in en-US
 ```
 
+## List Formatting
+
+Use `getSeparator` to get the appropriate list separator for the current locale. This is a separator that would be used in spoken language; note that the separator includes a space, for locales where it is appropriate.
+
+```javascript
+import {getSeparator} from '@brightspace-ui/intl/lib/list.js';
```
dlockhart, line 252: Just because our current lint rules will complain about this if people copy-paste:

```suggestion
import { getSeparator } from '@brightspace-ui/intl/lib/list.js';
```

test/list.js
```diff
@@ -0,0 +1,56 @@
...
+		{ locale: 'nl-NL', expect: [', ', ',\xa0'] },
+		{ locale: 'pt', expect: [', ', ',\xa0'] },
+		{ locale: 'pt-BR', expect: [', ', ',\xa0'] },
+		{ locale: 'sv', expect: [', ', ',\xa0'] },
+		{ locale: 'sv-SE', expect: [', ', ',\xa0'] },
+		{ locale: 'tr', expect: [', ', ',\xa0'] },
+		{ locale: 'tr-TR', expect: [', ', ',\xa0'] },
+		{ locale: 'zh', expect: ['、', '、'] },
+		{ locale: 'zh-CN', expect: ['、', '、'] },
+		{ locale: 'zh-TW', expect: ['、', '、'] },
```
dcostiuc, line 47: This is more of an aside (non-blocking side question), but how come we support [35 locales/variants here](https://github.com/BrightspaceUI/intl/blob/main/test/fileSize.js#L36-L70) but we only have [only 18 lang files](https://github.com/BrightspaceUI/core/tree/main/lang) in our core repo?

test/index.html
```diff
@@ -8,11 +8,13 @@
 	<div id="mocha"></div>
 		<script src="../node_modules/chai/chai.js"></script>
 		<script src="../node_modules/mocha/mocha.js"></script>
-		<script>mocha.setup('bdd')</script>
+		<script>mocha.setup('bdd');
+</script>
```
dlockhart, line 12: Was this intended?

dlockhart (approved):
Awesome!

dcostiuc (approved):
LGTM, but could probably use a review from others.

README.md
eKoopmans: Okay - I followed the formatting in the rest of the readme, should I update the spacing for all the imports?

Also, I think our readme is a bit confusing in its treatment of `options`, which is made to seem like an array - e.g.:

```js
const number = formatNumber(8902.72, [options]);
```

test/list.js
eKoopmans: FWIW, these tests helped me catch three cases I hadn't accounted for:

- `zh` (had only checked specifically `zh-cn` and `zh-tw`)
- `ar-sa`
- `ja-jp`

README.md
dlockhart: > should I update the spacing for all the imports?

Only if you want. Likely our linting rules have gotten a bit stricter since this was originally written.

> I think our readme is a bit confusing in its treatment of options

The square brackets were meant to mean "optional parameter", but since they also mean an array that's confusing for sure.

test/list.js
dlockhart: Yeah sometimes the base locale tests are handy because technically our clients can create whatever locale they want in the system and call it `en-zz` (or `fr-on`...) and this library should fall back gracefully.

dlockhart (approved):

test/list.js
eKoopmans: Hmm, so currently I've just hard-coded the special-case locales, including the base and sub-locales.

Should I instead check the base locale directly? I see that `fileSize` does:

```js
	const subtags = language.split('-');
	const baseLanguage = subtags[0];
```

dlockhart (approved):

PR #111, opened 2022-10-03
Remove eslint plugin dependencies.

dbatiste:
These dependencies should be installed as peer dependencies now.


dlockhart (approved):
Right -- good call. I've been lazy merging these upgrades and forgot that we could do this, thanks!

dbatiste:
Np! Yeah I'll probably just clean them up as I see them if I am working on something.

PR #112, opened 2022-11-25
feat: add locale names

dlockhart:
Exposing the localized locale names for our supported locales. This will make rendering a "language picker" dropdown easier.


dbatiste (approved):
Looks good. May want to do similar for the culture-specific locales depending on what you decide from this conv: https://github.com/BrightspaceUI/core/pull/3015#discussion_r1032627418

dlockhart:
Alright, I've switched this to expose details about the culture-specific locales instead. I don't see a need to expose both -- we can always add it later if it's needed.

PR #119, opened 2023-04-24
fix: Setup node version and update semantic-release

svanherk:
Updating to look more like our other repos, unless there was a reason we didn't specify a node version.  Going to do a minor release to make sure I didn't break that process.


dlockhart (approved):
LG -- can't think of a reason we weren't doing this before.

PR #124, opened 2023-09-08
feat: cache date/time descriptor

dlockhart:
We're [removing core's cache](https://github.com/BrightspaceUI/core/pull/3986) of the date/time descriptor because it had no way to know when to invalidate it when the language changed.

This re-adds the caching in a better place -- at the source!

### DIFF
--- a/lib/dateTime.js
+++ b/lib/dateTime.js
@@ -691,266 +691,267 @@ export function convertUTCToLocalDateTime(date) {
 }
 
 export function getDateTimeDescriptor() {
-
-	const language = getLanguage();
 	const settings = getDocumentLocaleSettings();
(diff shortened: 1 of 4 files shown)


lib/dateTime.js
```diff
@@ -691,266 +691,267 @@ export function convertUTCToLocalDateTime(date) {
 }
 
 export function getDateTimeDescriptor() {
```
dlockhart, line 693: Nothing changed here except everything go indented around the cache call.

```javascript
return settings.getCacheItem('dateTimeDescriptor', () => {
  // existing code here
});
```

dbatiste (approved):

PR #125, opened 2023-10-24
Add Maori support

bearfriend:
[US156955](https://rally1.rallydev.com/#/?detail=/userstory/720543634719&fdp=true): Intl

Adds Maori support to date/time, numbers, and supported lang/locales. Also moves some commonly repeated lang/locale lists into Intl, which can be imported from here going forward.

More list changes to come later.

### DIFF
--- a/lib/common.js
+++ b/lib/common.js
@@ -1,5 +1,5 @@
 export const defaultLocale = 'en';
-export const supportedBaseLocales = ['ar', 'cy', 'da', 'de', 'en', 'es', 'fr', 'hi', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh'];
+export const supportedBaseLocales = ['ar', 'cy', 'da', 'de', 'en', 'es', 'fr', 'hi', 'ja', 'ko', 'mi', 'nl', 'pt', 'sv', 'tr', 'zh'];
 export const supportedLocalesDetails = [
 	{ code: 'ar-sa', name: 'العربية' },
 	{ code: 'cy-gb', name: 'Cymraeg (Y Deyrnas Unedig)' },
@@ -16,6 +16,7 @@ export const supportedLocalesDetails = [
 	{ code: 'hi-in', name: 'हिन्दी Hindi (India)' },
 	{ code: 'ja-jp', name: '日本語 (日本)' },
 	{ code: 'ko-kr', name: '한국어 (대한민국)' },
+	{ code: 'mi-nz', name: 'Māori (Aotearoa)' },
 	{ code: 'nl-nl', name: 'Nederlands (Nederland)' },
 	{ code: 'pt-br', name: 'Português (Brasil)' },
 	{ code: 'sv-se', name: 'Svenska (Sverige)' },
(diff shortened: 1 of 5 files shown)


lib/common.js
```diff
@@ -25,6 +27,15 @@ export const supportedLocalesDetails = [
...
 export const supportedLocales = supportedLocalesDetails.map((l) => l.code);
 
+export const frauConfig = {
+	defaultLangTag: defaultLocale,
+	primary: supportedBaseLocales,
+	regions: supportedLocalesDetails.map(l => l.code.replace(/\w\w$/, r => r.toUpperCase())),
+	aliases: {
+		'en-AU': 'en-GB'
+	}
+};
```
dlockhart, line 37: What's this stuff for?

lib/common.js
```diff
@@ -25,6 +27,15 @@ export const supportedLocalesDetails = [
 ];
 export const supportedLocales = supportedLocalesDetails.map((l) => l.code);
```
dlockhart, line 28: Haha these different lists are pretty confusing and I was responsible for making some/all of them! Is it:

- `supportedLocales`: full base + regional codes... not sure what this is actually used for?
- `supportedLocalesDetails`: only used in core demo snippet
- `supportedBaseLocales`: supported locales but just the base part... seems to only be used within Intl
- `supportedLangpacks`: are these the full langpack codes coming out of Brightspace?

lib/common.js
bearfriend: This is to replace `frau-locale-provider` stuff.

lib/dateTime.js
```diff
@@ -825,6 +826,18 @@ export function getDateTimeDescriptor() {
...
+			case 'mi':
+				dateFormats = ['dddd, d MMMM y', 'd MMMM y', 'dd-MM-y', 'MMMM yyyy', 'd MMMM', 'd MMM'];
+				months = [
+					['Kohitātea', 'Huitanguru', 'Poutūterangi', 'Paengawhāwhā', 'Haratua', 'Pipiri', 'Hōngongoi', 'Hereturikōkā', 'Mahuru', 'Whiringa-ā-nuku', 'Whiringa-ā-rangi', 'Hakihea'],
+					['Kohi', 'Hui', 'Pou', 'Pae', 'Hara', 'Pipi', 'Hōngo', 'Here', 'Mahu', 'Nuku', 'Rangi', 'Haki']
+				];
+				days = [
+					['Rātapu', 'Rāhina', 'Rātū', 'Rāapa', 'Rāpare', 'Rāmere', 'Rāhoroi'],
+					['Tap', 'Hin', 'Tū', 'Apa', 'Par', 'Mer', 'Hor'],
+					['T', 'H', 'T', 'A', 'P', 'M', 'H']
```
bearfriend, line 838: These are now confirmed correct.

lib/common.js
dlockhart: Cool. Just wondering what value it has here vs. in `frau-localize-provider` directly and having `frau-locale-provider` import `defaultLocale`, `supportedBaseLocales` and `supportedLocalesDetails` from here. Just don't love this library needing to know about "frau" or "FRAs" at all.

bearfriend: Good point! I'll make that change.

dlockhart (approved):

PR #131, opened 2023-10-30
Add type to package.json

bearfriend:
There are some edge cases where imports from this library fail because it thinks it's a commonJS module. Frankly, I'm not sure why there hasn't been more issues without this declared.


dlockhart (approved):
Yep, agreed!

PR #179, opened 2024-07-17
fix: align with LMS naming

dlockhart:
This is to align things with what's in `LANG_LANGUAGES` in the LMS.


bearfriend:
[India in Hindi](https://github.com/unicode-org/cldr-json/blob/0876ec40e13d54c0a6b6456392802d4de7e059cb/cldr-json/cldr-localenames-full/main/hi/territories.json#L161)

[Saudi Arabia in Arabic](https://github.com/unicode-org/cldr-json/blob/0876ec40e13d54c0a6b6456392802d4de7e059cb/cldr-json/cldr-localenames-full/main/ar-SA/territories.json#L256)

bearfriend (approved):

PR #180, opened 2024-07-17
fix: update release timeout

dlockhart:
The previous release timed out because it took too long to add comments to all the PRs that were included (mostly Dependabot).  It didn't end up publishing to NPM.


bearfriend (approved):
😬

PR #186, opened 2024-08-22
Move to BrightspaceUI/testing

bearfriend:
While I was adding `Localize` tests I couldn't get the current setup to work at all, so I converted everything to `@brightspace-ui/testing`. This is slightly weird in that it depends on `intl`, but it's extremely minimal. It only [uses](https://github.com/BrightspaceUI/testing/blob/0e7b3421e66174fa8542f5a2592041724cda8036/src/browser/reset.js#L93) `documentLocaleSettings` once to check the language when resetting. I don't forsee it causing any real problems.


package.json
```diff
@@ -39,13 +39,8 @@
   "devDependencies": {
     "@babel/core": "^7",
     "@babel/eslint-parser": "^7",
```
dlockhart, line 41: I think `devDependencies` on `@babel/core`, `@babel/eslint-parser` can go away.

test/.eslintrc.json
dlockhart, line 1: Could this extend our shared testing config?

dlockhart (approved):

PR #189, opened 2024-08-23
Add `Localize` and `formatRelativeDateTime` docs

bearfriend:
The `Localize` docs largely pull from the existing `LocalizeMixin` [docs in core](https://github.com/BrightspaceUI/core/tree/main/mixins/localize#localizemixin), with some tweaks that we should copy back to there.

Also, apparently, we never documented `formatRelariveDateTime`


README.md
```diff
@@ -258,6 +268,103 @@ const separator = getSeparator({ nonBreaking: true }); // -> ',\xa0' in en-US
...
+
+### `localize()`
+
+The `localize()` method is used to localize a message.
+
+If the message contains arguments, provide replacement values in the second parameter:
+
+```javascript
+  const message = localizer.localize('hello', { firstName: 'Mary' });
+}
```
dlockhart, line 320:
```suggestion
const message = localizer.localize('hello', { firstName: 'Mary' });
```

dlockhart (approved):

lib/localize.js
bearfriend, line 1: Renamed this stuff just so autocomplete would look nicer

Screenshot 2024-08-23 at 9 12 08 AM

dlockhart: Yeah good call!

PR #192, opened 2024-08-26
Add locale direction

bearfriend:
Adds `dir` to `supportedLocaleDetails`. This can be used to correctly render individual tranlations on a page that otherwise uses a different locale.


dlockhart (approved):
Nice 👍

PR #200, opened 2024-09-24
Add PluralRules locale data shimming

bearfriend:
Support for Māori across browsers is inconsistent at best. Specifically for `Intl.PluralRules` (which is relied upon by `formatjs` to render `plural` arguments), no browsers have (proper) native support for Māori's plural categories. The `@formatjs/intl-pluralrules` polyfill also does not have data for Māori, and provides no value otherwise as `Intl.PluralRules` is well supported at this point.

We can rely on native `Intl.PluralRules` as long as we can shim locale data into it for `formatjs` to use. So, that's what we do here.

### DIFF
--- a/lib/PluralRules.js
+++ b/lib/PluralRules.js
@@ -49,7 +49,7 @@ class PluralRules extends Intl.PluralRules {
 
 	constructor(locales, options = {}) {
 		super(locales, options);
-		this.#locale = PluralRules.supportedLocalesOf(getCanonicalLocales(locales))[0];
+		this.#locale = PluralRules.supportedLocalesOf(locales)[0];
 		this.#type = options.type ?? 'cardinal';
 		if (localeData[this.#locale]) {
 			this.#localeData = localeData[this.#locale];
--- a/lib/PluralRules.js
+++ b/lib/PluralRules.js
@@ -1,0 +1,70 @@
+//import '@formatjs/intl-pluralrules/polyfill.js';
+
+const customLocales = ['mi'];
+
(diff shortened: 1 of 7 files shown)


lib/PluralRules.js
```diff
@@ -0,0 +1,78 @@
...
+			},
+		},
+		select(n, ord) {
+			return !ord && n === 1 ? 'one' : 'other';
+		}
+	}
+};
+
+function getCanonicalLocales(locales) {
+	const mappedLocales = [locales].flat().map(locale => {
```
bearfriend, line 23: This is only needed for Safari on older OS versions (< ~macOS 12, based on sauce testing), which don't map the aliases back to `mi`.

package.json
```diff
@@ -44,7 +44,6 @@
     "sinon": "^19.0.2"
   },
   "dependencies": {
-    "@formatjs/intl-pluralrules": "^1",
```
dlockhart, line 47: Just to clarify: the other things that this polyfill was providing (the plural rules locale data for browsers who don't support the native API) aren't needed at all anymore?

bearfriend: Ack! Safari 12 is the only browser that isn't blocked that doesn't support `Intl.Pluralules`. Do we want to keep it for that?

bearfriend: Discussed offline.

Brightspace loads... something in Safari 12 but it's completely unuseable so we're going ahead with removing the polyfill. We expect to block Safari 12 in the near future.

dlockhart (approved):

PR #208, opened 2024-11-08
feat: add ability to localize common characters

dlockhart:
I had some trouble figuring out how best to integrate this, so definitely interested in feedback and other ideas.

The big downside to this approach is that every component that wires up to a localizer via the `LocalizeMixin` is going to have these "common" terms merged into their collection. That's why I needed to use the `intl-common` prefix on all their keys to avoid a potential collission.

~It also means that the common terms will **always** be fetched once on every page, even if they're not used. Both `localize` and `localizeHTML` are synchronous, so I couldn't think of a way to defer the loading of the common stuff.~ That's no longer true now that `loadCommon: true` must be set in the config.

### DIFF
--- a/lang/en.js
+++ b/lang/en.js
@@ -1,15 +1,15 @@
 export default {
-	"intl-common:characters:apostrophe": "apostrophe", // description of the "'" character
-	"intl-common:characters:ampersand": "ampersand", // description of the "&" character
-	"intl-common:characters:asterisk": "asterisk", // description of the "*" character
-	"intl-common:characters:backslash": "backslash", // description of the "\" character
-	"intl-common:characters:colon": "colon", // description of the ":" character
-	"intl-common:characters:comma": "comma", // description of the "," character
-	"intl-common:characters:greaterThan": "greater-than sign", // description of the ">" character
-	"intl-common:characters:lessThan": "less-than sign", // description of the "<" character
-	"intl-common:characters:numberSign": "number sign", // description of the "#" character
-	"intl-common:characters:percentSign": "percent sign", // description of the "%" character
-	"intl-common:characters:pipe": "pipe", // description of the "|" character
-	"intl-common:characters:questionMark": "question mark", // description of the "?" character
-	"intl-common:characters:quotationMark": "quotation mark", // description of the '"' character
+	"intl-common:characters:apostrophe": "apostrophe", // short name or description of the "'" character
+	"intl-common:characters:ampersand": "ampersand", // short name or description of the "&" character
+	"intl-common:characters:asterisk": "asterisk", // short name or description of the "*" character
+	"intl-common:characters:backslash": "backslash", // short name or description of the "\" character
+	"intl-common:characters:colon": "colon", // short name or description of the ":" character
+	"intl-common:characters:comma": "comma", // short name or description of the "," character
+	"intl-common:characters:greaterThan": "greater-than sign", // short name or description of the ">" character
+	"intl-common:characters:lessThan": "less-than sign", // short name or description of the "<" character
+	"intl-common:characters:numberSign": "number sign", // short name or description of the "#" character
+	"intl-common:characters:percentSign": "percent sign", // short name or description of the "%" character
+	"intl-common:characters:pipe": "pipe", // short name or description of the "|" character
+	"intl-common:characters:questionMark": "question mark", // short name or description of the "?" character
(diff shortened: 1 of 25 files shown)


lang/en.js
dlockhart, line 1: Once we settle on a solution, I'll fill in the rest of the translations.

dbatiste:
> The big downside to this approach is that every component that wires up to a localizer via the `LocalizeMixin` is going to have these "common" terms merged into their collection. That's why I needed to use the `intl-common` prefix on all their keys to avoid a potential collission.

If more than one collection is fetched, does that mean they are effectively multiple copied of it in memory? It's probably not going to be a huge set, but doesn't seem ideal. Maybe there is a way to share one set? Or, maybe there could be a way for those components to indicate whether to include the common terms in the set? I'm not too familiar with how the terms are fetched, so maybe none of those things are possible. 

> It also means that the common terms will **always** be fetched once on every page, even if they're not used. Both `localize` and `localizeHTML` are synchronous, so I couldn't think of a way to defer the loading of the common stuff.

I think async `localize` could potentially complicate things quite a bit for consumers.

lib/localize.js
```diff
@@ -166,6 +189,9 @@ export const getLocalizeClass = (superclass = class {}) => class LocalizeClass e
 			const possibleLanguages = this._generatePossibleLanguages(config);
 			const resourcesPromise = this.getLocalizeResources(possibleLanguages, config);
 			resourcesLoadedPromises.push(resourcesPromise);
+			resourcesLoadedPromises.push(this.getLocalizeResources(possibleLanguages, {
```
dlockhart, line 192: I just realized I can maybe leverage the `localizeConfig` to signal whether or not to load the common stuff -- I'll give that a try and see how it looks.

lib/localize.js
```diff
@@ -164,8 +191,15 @@ export const getLocalizeClass = (superclass = class {}) => class LocalizeClass e
 		}
 		if (Object.prototype.hasOwnProperty.call(this, 'getLocalizeResources') || Object.prototype.hasOwnProperty.call(this, 'resources')) {
 			const possibleLanguages = this._generatePossibleLanguages(config);
-			const resourcesPromise = this.getLocalizeResources(possibleLanguages, config);
-			resourcesLoadedPromises.push(resourcesPromise);
+			if (config?.importFunc) {
```
dlockhart, line 194: I had to add this check to handle the case where the consumer _only_ wants to use the common resources.

lib/localize.js
```diff
@@ -164,8 +191,15 @@ export const getLocalizeClass = (superclass = class {}) => class LocalizeClass e
...
+				const resourcesPromise = this.getLocalizeResources(possibleLanguages, config);
+				resourcesLoadedPromises.push(resourcesPromise);
+			}
+			if (config?.loadCommon) {
```
dlockhart, line 198: This is now only loading the common resources if the consumer explicitly asks for it.

bearfriend:
> If more than one collection is fetched, does that mean they are effectively multiple copies of it in memory?

I think that might be true currently, but we should be able to shift things around a bit to avoid that. The set might actually grow quite a bit so we should be sure about that.

Otherwise this all looks really good to me.

lib/localize.js
```diff
@@ -164,8 +193,21 @@ export const getLocalizeClass = (superclass = class {}) => class LocalizeClass e
...
+				resourcesLoadedPromises.push(this.getLocalizeResources(possibleLanguages, {
+					importFunc: async(lang) => {
+						if (commonResources.has(lang)) return commonResources.get(lang);
+						const resources = (await import(`../lang/${lang}.js`)).default;
+						commonResourcesImportCount++;
```
dlockhart, line 205: This now keeps a local cache of the common resources, which prevents multiple in-memory copies of the same resources for a given language.

Testing that this was actually working was tricky since we can't observe whether a dynamic import happens or not. I went with a simple counter, which is maybe a bit fragile but gets the job done.

bearfriend: This is good for now, I think, but the problem is that we make copies of everything by combining them into a single object per-element, which is bad. We should make a separate story to address this before we expand the list of common messages.

lang/en.js
bearfriend: I think we'll need to add context with each actual character to these

dbatiste (approved):

PR #212, opened 2024-11-18
Localize common characters (attempt 2)

dlockhart:
I had to quickly revert the [first version](https://github.com/BrightspaceUI/intl/pull/208) of this as it broke a bunch of core unit tests. The reason for that is that the changes caused components wired up to use a `static` collection of resources (i.e. no `importFunc`) to stop working.

I'll add comments to call out the changes I made to work around this.


lib/localize.js
```diff
@@ -129,12 +158,25 @@ export const getLocalizeClass = (superclass = class {}) => class LocalizeClass e
 			const possibleLanguages = this._generatePossibleLanguages(config);
```
dlockhart, line 158: Change 1: these lines were previously wrapped around an `if (config.importFunc) {` but components that define their own static resources by manually implementing `getLocalizeResources` were then not getting their resources filled in.

lib/localize.js
```diff
@@ -129,12 +158,25 @@ export const getLocalizeClass = (superclass = class {}) => class LocalizeClass e
...
+					}
+				}));
+			}
 		}
 		return Promise.all(resourcesLoadedPromises);
 	}
 
 	static async _getLocalizeResources(langs, { importFunc, osloCollection, useBrowserLangs }) {
 
+		if (importFunc === undefined) return;
```
dlockhart, line 178: Change 2: the default implementation of `getLocalizeResources` just calls this, but `importFunc` can now be `undefined`. So we short-circuit and return `undefined`, but that's going to cause change 3 below...

lib/localize.js
```diff
@@ -177,6 +219,13 @@ export const getLocalizeClass = (superclass = class {}) => class LocalizeClass e
 
 		const allResources = {};
 		const resolvedLocales = new Set();
+
+		localizeResources.forEach((entry, index) => {
```
dlockhart, line 223: Change 3: because of change 2, some of these resources can be `undefined` -- so we filter them out.

dbatiste (approved):

lib/localize.js
bearfriend: No idea about the efficiency but could this be a filter or reduce instead?

bearfriend: Like the `flat` above could probably just be a reduce that does both.

bearfriend (approved):

PR #227, opened 2025-01-13
Add Hawaiian to supported languages

bearfriend:
[GAUD-7412](https://desire2learn.atlassian.net/browse/GAUD-7412): Add `haw` as supported language in `brightspace-ui/intl`

Adds `haw`/`haw-us` as a supported lang/locale. More data for dates/times, numbers etc. will be added separately.

[GAUD-7412]: https://desire2learn.atlassian.net/browse/GAUD-7412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/common.js
+++ b/lib/common.js
@@ -1,6 +1,6 @@
 export const defaultLocale = 'en';
-export const supportedBaseLocales = ['ar', 'cy', 'da', 'de', 'en', 'es', 'fr', 'hi', 'ja', 'ko', 'mi', 'nl', 'pt', 'sv', 'tr', 'zh'];
-export const supportedLangpacks = ['ar', 'cy', 'da', 'de', 'en', 'en-gb', 'es', 'es-es', 'fr', 'fr-fr', 'fr-on', 'hi', 'ja', 'ko', 'mi', 'nl', 'pt', 'sv', 'tr', 'zh-cn', 'zh-tw'];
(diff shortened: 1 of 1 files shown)


lib/common.js
```diff
@@ -14,6 +14,7 @@ export const supportedLocalesDetails = [
 	{ code: 'fr-ca', dir: 'ltr', name: 'Français (Canada)' },
 	{ code: 'fr-fr', dir: 'ltr', name: 'Français (France)' },
 	{ code: 'fr-on', dir: 'ltr', name: 'Français (Ontario)' },
+	{ code: 'haw-us', dir: 'ltr', name: 'ʻŌlelo Hawaiʻi (ʻAmelika Hui Pū ʻIa)' },
```
KearseTrevor, line 17: is `-us` necessary for haw?

KearseTrevor (approved):

lib/common.js
KearseTrevor: Specifically, instead of `haw-haw`

bearfriend: Regions must be a valid [ISO 3166](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes) code. It's arguable whether using them at all for some languages is necessary, but I'm uncertain if leaving it off breaks stuff in the rest of the system. We may use it for figuring out which number system to use for example in the LMS so for now we just keep it consistent.

PR #238, opened 2025-03-05
Add a datalist of supported locales using their local names

bearfriend:
Adds a datalist to the lang attribute input on the demo page to make finding supported locales easier/searchable while still allowing "arbitrary" codes (e.g. `fr-be`, `xyz`)

Also adds local names to `supportedLocalesDetails` that will remain undocumented for now. This sets up some future work to transition to browser internationalization APIs and standard codes that either better represent our locales or are better recognized by APIs.

Screenshot 2025-03-05 at 12 33 40 PM

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -28,7 +28,7 @@
 		window.intl = { ...common, ...dateTime, ...fileSize, ...list, ...number, ...localize };
 
 		render(
-			common.supportedLocalesDetails.map(({ code, name }) => html`<option value="${code}">${name}</option>`),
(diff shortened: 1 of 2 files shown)


demo/index.html
```diff
@@ -49,6 +57,7 @@ <h2>Usage</h2>
 		const htmlLangInput = document.getElementById('html-lang');
 
 		document.getElementById('apply-lang').onclick = () => {
+			debugger;
```
dlockhart, line 60: Did you mean to leave this in?

bearfriend: Hah, nope

dlockhart (approved):

dbatiste (approved):

PR #239, opened 2025-03-10
Filter out bad locales and fail gracefully

bearfriend:
[GAUD-7600](https://desire2learn.atlassian.net/browse/GAUD-7600): intl: Filter bad locales and fail gracefully

In addition to the invalid `en-US@posix` locale, some environments have an empty string. This fix is in addition to [setting the `LANG` env var](https://github.com/BrightspaceUI/testing/pull/630/files), which is preferable to these fallbacks and failures.

- Fall back to `defaultLocale`
- Filter out [invalid locales](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locales_argument)
- Fail gracefully via try-catch

[GAUD-7600]: https://desire2learn.atlassian.net/browse/GAUD-7600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/common.js
+++ b/lib/common.js
@@ -93,7 +93,7 @@ class DocumentLocaleSettings {
 
 	constructor() {
 		this._cache = new Map();
-		this._htmlElem = window.document.getElementsByTagName('html')[0];
+		this._htmlElem = document.documentElement;
 		this._listeners = [];
 		this._overrides = {};
 		this._observer = new MutationObserver(this._handleObserverChange.bind(this));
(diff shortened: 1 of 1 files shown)


nathankoop:
Can we add some tests to this code?
Not needing it to block this merge, but as a task for us to do tomorrow?

PR #248, opened 2025-04-24
Add pseudolocalization

bearfriend:
[GAUD-7785](https://desire2learn.atlassian.net/browse/GAUD-7785): Make pseudo-l10n work in our Internationalization (i18n) methods

Extends [Pseudolocalization](https://desire2learn.atlassian.net/wiki/spaces/DEVCENTRAL/pages/3345055961/Pseudo-localization) to anything using this library i.e. everything offstack.

The goal was to not clutter each `localize` call with a check since it is called so frequently. Instead we set the method differently at construction time.

We also move `DocumentLocaleSettings` out to its own file and clean it up a bit.

[GAUD-7785]: https://desire2learn.atlassian.net/browse/GAUD-7785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/document-locale-settings.js
```diff
@@ -0,0 +1,140 @@
+const host = document.documentElement;
+
+const DEFAULTS = {
+	pseudoLocalization: { isAvailable: false, isEnabledForUser: false, textFormat: '{0}' },
```
dlockhart, line 4: Do we actually need to send down `isAvailable` and `isEnabledForUser`, or could we just not even set the `data-` attribute if both those things aren't true? Like those are server-side things that it doesn't feel like the front end needs to know about...

lib/localize.js
```diff
@@ -35,6 +35,13 @@ const noAllowedTagsRegex = getDisallowedTagsRegex([]);
 export const getLocalizeClass = (superclass = class {}) => class LocalizeClass extends superclass {
 
 	static documentLocaleSettings = getDocumentLocaleSettings();
+
+	static pseudoLocalize(localize, name, ...replacements) {
```
dlockhart, line 39: TODO: tests?

lib/document-locale-settings.js
bearfriend: Yeah how we do that is totally up in the air. What you suggest makes sense to me. I mostly just put everything in here that I think we ultimately _might_ need (even if only on the server). So, I didn't include any of the character look-alike stuff etc.

lib/localize.js
```diff
@@ -258,7 +265,9 @@ export const Localize = class extends getLocalizeClass() {
 	constructor(config) {
 		super();
 		super.constructor.setLocalizeMarkup(localizeMarkup);
-		this.localize = (...args) => super.localize(...args);
+		this.localize = super.constructor.documentLocaleSettings.pseudoLocalization?.isAvailable
+			? (...args) => super.constructor.pseudoLocalize((...args) => super.localize(...args), ...args)
```
bearfriend, line 269: Somewhat surprisingly, this is appreciably faster than various other ways to bind `this` and pass the arguments through.

test/localize.test.js
```diff
@@ -140,6 +158,34 @@ describe('Localize', () => {
...
+		afterEach(() => {
+			pseudoLocalizeSpy.restore();
+		});
+
+		it('should not run by default', async() => {
+			localizer = new Localize({ importFunc: lang => resources[lang] });
+			await localizer.ready;
+			localizer.localize('basic', { employerName: 'D2L' });
+			expect(pseudoLocalizeSpy).to.not.have.been.called;
+		})
```
bearfriend, line 177:
```suggestion
		});
```

dlockhart (approved):
Looks good!

PR #257, opened 2025-06-23
GAUD-8233: deep copy defaults upon reset

dlockhart:
Fixing a regression from #248.

For better or worse, `localize-behavior` is [directly modifying](https://github.com/BrightspaceUI/localize-behavior/blob/main/test/d2l-localize-behavior.test.js#L326) the locale setting's timezone name and identifier. Because those are both just pointers to `DEFAULTS`'s references, in doing so it's actually changing the `DEFAULTS`. Then when `reset()` does an `Object.assign` they propagate.

The fix here is to do a deep copy of `DEFAULTS` during `reset()`.

This will fix the [failing CI in localize-behavior](https://github.com/BrightspaceUI/localize-behavior/pull/67).

### DIFF
--- a/lib/document-locale-settings.js
+++ b/lib/document-locale-settings.js
@@ -75,7 +75,7 @@ export class DocumentLocaleSettings {
 		this._cache.clear();
 		this._language = this._languageInitial;
 		this._listeners = [];
-		Object.assign(this, DEFAULTS);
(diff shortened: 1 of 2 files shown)


dbatiste (approved):

bearfriend (approved):
Ugh, no it won't freeze deeply.

PR #262, opened 2025-07-22
NTNGL-4668: introduce localizeCommon

dlockhart:
This adds something we've talked about for a while: a `localizeCommon(name)` for terms that show up a lot.

The motivation for doing this now is that Insights just added yet another usage of the term "Back" for the short button text in their immersive navigation usage. I copied the translations from the LMS's `Standard.Buttons.btnBack` term.

If we like the approach here, I'm happy to add a few more related "navigation" terms (Forward, Next, Previous, any others?).

### DIFF
--- a/lib/localize.js
+++ b/lib/localize.js
@@ -20,6 +20,7 @@ const characterMap = new Map([
 	['?', 'questionMark'],
 	['"', 'quotationMark']
 ]);
+const commonTerms = ['navigation:back:title'];
 const commonResources = new Map();
 export let commonResourcesImportCount = 0;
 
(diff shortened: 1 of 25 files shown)


lib/localize.js
```diff
@@ -20,6 +20,7 @@ const characterMap = new Map([
 	['?', 'questionMark'],
 	['"', 'quotationMark']
 ]);
+const commonTerms = ['navigation:back:title'];
```
dlockhart, line 23: I grouped this under "navigation" but happy to discuss other ways to group things!

dbatiste: This is great. I've noticed a lot of repeated terms elsewhere for workflow buttons (like `Cancel`, `Save`). I'm guessing they would be something like `button.save`, `button.cancel`? Not sure if it should be plural or not.

dlockhart: I don't know if I'd want to tie "button" to it -- which implies the component to use -- but maybe something like "action" or "verb"? That way "Save", "Cancel", "Save & New", "Delete", "Copy" could all kind of fit in there. 🤷

dlockhart: But then you could argue that "Back" is also an action/verb, so maybe not having a category at all is better?

NicholasHallman:
Awesome! As the number of terms build we'll likely want some documentation like the icon library that tells you which terms are available (future problems though!)

lib/localize.js
dbatiste: Sounds good to me.

dbatiste (approved):

PR #273, opened 2025-09-02
Add timezone data and tools

bearfriend:
[GAUD-8593](https://desire2learn.atlassian.net/browse/GAUD-8593): Create a standardized method in BrightspaceUI/intl to get and/or filter supported timezone data

- This is definitely not a complete or super clean implementation, though it's not terrible. We're mostly just looking to ship the main `lib/timeZones.js` for consumer use. The rest backing that up will almost certainly change at least a bit.
- Firefox does not support `Intl.Locale.prototype.getTimeZones()` so we need to map our own, unfortunately
- Once the additional requirements are fleshed out we can come back and clean up everything in `_timeZones/`, add more specific tests, docs, and make more of that available to consumers as well
- I'll be testing this in `core` to back up `d2l-input-time` custom timezone support, which @GZolla has a draft up for [here](https://github.com/BrightspaceUI/core/pull/6081)

[GAUD-8593]: https://desire2learn.atlassian.net/browse/GAUD-8593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/timeZones.js
```diff
@@ -0,0 +1,48 @@
...
+			timeZones = mapper.getTimeZonesForRegion(region);
+		}
+	} else {
+		if (timeZoneData) return timeZoneData;
+		timeZones = timeZoneIdentifiers;
+	}
+	return timeZones.map(l => getTimeZoneData(l));
+}
+
+export function getTimeZoneData(identifier = getDocumentLocaleSettings().timeZone.identifier, locale = getDocumentLocaleSettings().language || 'en') {
```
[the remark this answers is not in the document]
bearfriend: So currently we get an error with an invalid timezone from the `Intl` API, which I think is what we want here instead of a silent fail:
```
Uncaught RangeError: Invalid time zone specified: America/New_Yorkk
```

In the input itself, we just silently set `time-zone-hidden`, but I'm open to other strategies there.

lib/_timeZones/data/index.js
svanherk, line 1: Are all these pulled from the CLDR and part of what we hope to automate?

lib/_timeZones/mapper.js
svanherk, line 1: So this is only used for Firefox? Seems like no, `getRegionName` is used from here regardless. Was this grabbed from somewhere?

lib/_timeZones/mapper.js
```diff
@@ -0,0 +1,236 @@
+/**
+ * TimeZone Utilities
+ *
+ * This module provides utilities for working with IANA timeZone identifiers
+ * and their corresponding region information using ES modules.
+ *
+ * Data is based on the IANA Time Zone Database (tzdata) and excludes
+ * backward compatibility and etcetera entries.
```
svanherk, line 8: Aside, what are etcetera entries?

svanherk (commented):
I know this is merged, but I had started reviewing lol, so going to post it anyway

PR #358, opened 2026-04-21
Initial script to build locale data

bearfriend:
[GAUD-9714](https://desire2learn.atlassian.net/browse/GAUD-9714): intl: Create a build-locale-data script based on the mfv build script

This stuff is going to get moved around and refactored a bunch so this is just the initial script and resulting files, which for now will just sit in the repo without being bundled/shipped

[GAUD-9714]: https://desire2learn.atlassian.net/browse/GAUD-9714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mfv.config.json
```diff
@@ -0,0 +1,14 @@
...
+  "localesMap": {
+    "ar": "ar-eg",
+    "es-es": "es",
+    "es": "es-mx",
+    "fr-fr": "fr",
+    "fr-on": "fr-ca",
+    "fr": "fr-ca",
+    "zh-cn": "zh-Hans",
+    "zh-tw": "zh-Hant"
+  }
```
bearfriend, line 13: Because our locales are wonky, we need to map them to the source we want to actually pull from so that, for example, when a user requests `fr`, they get `fr-CA` date formatting instead of for the CLDR likely subtag `fr-FR`.

bearfriend: The reason this is in the `mfv` config is because the same map will start to be used when building locale data in the `t9n-format` action, or if you run `mfv format` locally. Right now its default data is just accommodating our weirdness.

locale-data/locales/en.js
bearfriend, line 1: We absolutely do not need all of this data, and we may need some things not yet here. I'll refine this as I rewrite more of the `lib/` files.

locale-data/build/utils.js
bearfriend, line 1: This is some AI magic that works but I don't entirely trust. I'll review this as well and rewrite it less... AI-y

locale-data/build/build-locale-data.js
```diff
@@ -0,0 +1,85 @@
...
+
+	const locales = config.path && (await readdir(`${cwd()}/${config.path}`).catch(() => {}))?.map(f => f.split('.')[0]);
+	const padLength = Math.max(...locales.map(l => l.length)) + 1;
+	locales.forEach(locale => {
+		const originalLocale = locale;
+		try {
+			locale = config.localesMap[locale] || locale;
+			locale = Intl.getCanonicalLocales(locale.trim().toLowerCase())[0];
+			stdout.write(`${originalLocale.padEnd(padLength)}${locale !== originalLocale ? ` ->  ${locale}` : ''}\n`);
+		} catch (e) {
```
dlockhart, line 43: Is it not possible to avoid the exception?

locale-data/build/utils.js
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Capitalizes the first letter of a month name if CLDR data indicates it should
+ * be title-cased in a UI list or menu context for the given locale.
+ *
+ * @param {string} monthName - The month name to potentially capitalize
+ * @param {string} locale - BCP 47 locale tag (e.g. 'es', 'cs', 'fr-CA')
+ * @returns {string}
+ */
+export async function capitalizeMonthForUI(monthName, locale) {
```
dlockhart: I don't think this is used (yet)? Is the idea that it'd be used in `dateTime.js`, or should it maybe live there if consumers will be importing it?

locale-data/build/utils.js
```diff
@@ -0,0 +1,89 @@
...
+ * that month names should be title-cased, for each combination of:
+ *   - grammatical form: "format" (used within a date) vs "standalone" (used alone)
+ *   - display context: "standAlone" (plain text isolation) vs "uiListOrMenu" (e.g. <select>)
+ *
+ * Checks src/locale-data-overrides/<locale>.js before falling back to CLDR XML.
+ *
+ * @param {string} locale - BCP 47 locale tag (e.g. 'es', 'cs', 'fr-CA')
+ * @returns {{ formatStandAlone: boolean, formatUiListOrMenu: boolean, standaloneStandAlone: boolean, standaloneUiListOrMenu: boolean }}
+ */
+export async function shouldTitleCaseMonths(locale) {
```
dlockhart, line 20: I'm a bit confused as to why the month names just wouldn't be capitalized automatically if they're supposed to be?

locale-data/build/utils.js
```diff
@@ -0,0 +1,89 @@
...
+	};
+
+	const candidates = [
+		locale.replace('-', '_'),
+		locale.split('-')[0],
+	];
+
+	for (const candidate of candidates) {
+		try {
+			const override = await import(`${overridesPath}/${candidate}.js`);
```
dlockhart, line 34: I guess these files will come later?

mfv.config.json
dlockhart: Does that mean every repo is going to have to have this config, or can it be shared somehow?

locale-data/locales/en.js
```diff
@@ -0,0 +1,1271 @@
...
+			"other"
+		],
+		ordinal: [
+			"one",
+			"two",
+			"few",
+			"other"
+		]
+	},
+	langaugeDisplayNames: {
```
dlockhart: Haha yeah hopefully we can slim some of this down so we don't use up the whole of the Internet.

locale-data/build/utils.js
bearfriend: Yeah this is useless right now. I just didn't bother deleting it until I go over the whole file. But yes it'll be used when formating dates and likely in a function to get months specifically, with a of couple required flags for usage details

locale-data/build/utils.js
bearfriend: Oh whoops I do have one for Catalan I'll add

locale-data/build/utils.js
bearfriend: You mean like already in the static CLDR data, or by `Intl`?

There are different rules for different usages. Some languages don't capitalize month names in date formatting or sentence form, so they're lowercase in CLDR, but if it's standalone in a calendar header or in a select list they do.

locale-data/build/build-locale-data.js
bearfriend: Like to just keep going if one fails? I was thinking we don't want to not notice it and accidentally delete a locale if something goes wrong.

locale-data/build/utils.js
dlockhart: Ah ok that makes sense!

locale-data/build/build-locale-data.js
dlockhart: I guess I'm not sure what in here could throw? Is it the `getCanonicalLocales` call, or referencing `[0]` if the array is empty? Or... ?

mfv.config.json
bearfriend: Still considering our options here. The action can set an environment variable, and most people aren't running `mfv format` locally, but I'd still like to come up with something cleaner.

locale-data/build/build-locale-data.js
bearfriend: Oh, I see what happened. There used to be a `writeFile` in the `try` that got moved. So, yeah I can get rid of it.

dlockhart (approved):

PR #376, opened 2026-06-10
Build LMS intl files

bearfriend:
[GAUD-9716](https://desire2learn.atlassian.net/browse/GAUD-9716): Build XML for LMS

Definitely some cleanup that is more obvious in the github diff.

- [ ] Remove `ru` for now
- [ ] Lint
- [ ] Remove `build-core-files`
- [ ] Clean up minor AI-generated verbosity
- [ ] Revert and regenerate `supported.js`

[GAUD-9716]: https://desire2learn.atlassian.net/browse/GAUD-9716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


build-locale-data/lms/LANG_LANGUAGES.xml
dlockhart, line 1: So is the idea that these XML table definitions will be generated here, included in the npm package and then copied (manually?) into the LMS repo?

bearfriend: For now, yeah. We can probably do a quick follow-up to automatically open PRs. All the existing data is going to be stale until next year when we can update it and then start keeping it up to date. Maybe once we do that we can switch to to automatic PRs.

PR #378, opened 2026-06-10
Rewrite dateTime.js to use current localeData

bearfriend:
[GAUD-9717](https://desire2learn.atlassian.net/browse/GAUD-9717): Rewrite intl/common and intl/dateTime to use contained data sets per locale

[GAUD-9717]: https://desire2learn.atlassian.net/browse/GAUD-9717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### [Run #1](https://github.com/BrightspaceUI/intl/actions/runs/28829793692/job/85500951715?pr=378)
- Use new `locale-data/`
- Do not modify tests
### [Run #2](https://github.com/BrightspaceUI/intl/actions/runs/28833593053/job/85512580980?pr=378)
- Refactor tests so they properly set the language between tests, including waiting for locale data to load
- Fix `documentLocaleSettings` `reset` method so it runs listeners when it "resets" (changes) the language back to English. This is only used in tests.
- Rename `locale-data/` to match the CLDR source names and update how the files are matched for loading
### [Run #3](https://github.com/BrightspaceUI/intl/actions/runs/28834235277/job/85514549154?pr=378)
- Update tests with appropriate whitespace (no-break-space, rtl-mark etc.) to match new locale data. These changes will only improve UX without being disruptive.
### [Run #4](https://github.com/BrightspaceUI/intl/actions/runs/28835162222/job/85517281938?pr=378)
- Be more forgiving when relying on separators to parse dates and date formats since they may contain invisible whitespace (e.g. rtl-mark)
- Support unicode symbols without a .NET equivalent
  - `LLLL` - Stand-alone month in year (e.g. Catalan: `MMMM` -> "d'abril", `LLLL` -> "abril")
  - `cccc` - Stand-alone day of week
  - `K` - Hour in day, 12 hour, 0-11 (e.g. "0:29 pm")
### [Run #5](https://github.com/BrightspaceUI/intl/actions/runs/28839577292/job/85530586792)
- Downgrade `locale-data/` files to maintain existing time formats (minus whitespace improvements and legitimate bugs)
- Where CLDR is wrong (`ko`), add `overrides` files and rework how they are applied
### Run #6
- Downgrade `locale-data/` files to maintain existing date formats (minus whitespace improvements and legitimate bugs)
- Align `pt` and `pt-BR`
#### 0 failed tests


lib/dateTime.js
```diff
@@ -29,14 +26,14 @@ const timezoneOffsetMap = {
 	'AZST': '+0500',
 	'AZT': '+0400',
 	'B': '+0200',
-	'BDST': '+0100',
+	'BDST': '+0200',
```
dlockhart: These were just mistakes you noticed?

lib/dateTime.js
```diff
@@ -692,295 +661,241 @@ export function convertUTCToLocalDateTime(date) {
 	return convertJsDateToLocalDateTime(utcDate) || date;
 }
 
+function convertCldrDatePattern(pattern, forceFullYear) {
```
dlockhart: Maybe some unit tests for this guy?

lib/dateTime.js
```diff
@@ -692,295 +661,241 @@ export function convertUTCToLocalDateTime(date) {
...
+		const fullTimeFormat = tzFirst ? `ZZZ ${timeFormat}` : `${timeFormat} ZZZ`;
+
+		// Convert date formats from CLDR
+		const dateFormats = {
+			'full': convertCldrDatePattern(localeData.dateFormats.full, false),
+			'medium': convertCldrDatePattern(localeData.dateFormats.medium, false),
+			'short': convertCldrDatePattern(localeData.dateFormatItems.yMd, true),
+			'monthYear': convertCldrDatePattern(localeData.dateFormatItems.yMMMM, false),
+			'monthDay': convertCldrDatePattern(localeData.dateFormatItems.MMMMd, false),
+			'shortMonthDay': convertCldrDatePattern(localeData.dateFormatItems.MMMd, false),
```
dlockhart: This stuff ends up all being static eh? Wonder if it'd be useful doing these calculations at build-time instead of runtime and storing them in `locale-data.js`?

lib/dateTime.js
```diff
@@ -692,295 +661,241 @@ export function convertUTCToLocalDateTime(date) {
...
-		let days = [
-			['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'],
-			['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'],
-			['S', 'M', 'T', 'W', 'T', 'F', 'S']
-		];
-		let firstDayOfWeek = (mondayFirstDayLocales.indexOf(baseLanguage) > -1) ? 1 : 0;
-		let weekendStartDay = 6;
-		let weekendEndDay = 0;
-
-		switch (baseLanguage) {
```
dlockhart, line 729: Nice! The fact that all the unit tests passed seems to indicate that the CLDR data was a match for what we were doing here?

lib/dateTime.js
bearfriend: "Mistakes" AI noticed, but I don't trust it understands the context. I'll make a story to review these, but will revert these from this PR.

lib/dateTime.js
bearfriend: I'll look at this, but it's cached after the first run. I do want to avoid storing converted data in `locale-data/` so we just have a clean source of truth. A lot of this may go away very soon too with the skeleton work coming up, so we can look at it again then.

lib/dateTime.js
bearfriend: If only. I had a bad import reference so basically none of the ones that matter ran.

test/dateTime.test.js
```diff
@@ -354,14 +371,14 @@ describe('dateTime', () => {
 			['early', 'late'].forEach((timeOfDay) => {
 				['short', 'full'].forEach((format) => {
 					[true, false].forEach((hour24) => {
-						it(`should format ${input.locale}/${timeOfDay}/${format}/${hour24}`, () => {
-							documentLocaleSettings.language = input.locale;
+						index++;
+						((idx) => it(`should format ${input.locale}/${timeOfDay}/${format}/${hour24}`, async() => {
+							await setLanguage(input.locale);
 							documentLocaleSettings.overrides = { date: { hour24: hour24 } };
-							index++;
```
bearfriend, line 360: These old `index++`'s only work if you run all the tests. Using `--grep` breaks them since the indexes no longer match.

lib/dateTime.js
bearfriend: Like it seems pretty well covered by all the existing (and new) output tests, but do you think we should dump/reduce those in favor of testing things like this more directly? It's really not about individual locales anymore so much as all the mapping.

Again, lots of this may go away soon. Maybe we rely on the existing tests for now and reassess when we add skeleton support, which I should have a clearer picture of by EOD tomorrow.

build-locale-data/locale-data-overrides/ko.js
bearfriend, line 1: This is the only case of finding the CLDR to be wrong/outdated. For that, we use the overrides functionality. For everything _we're_ wrong about, we overwrite the values into the `locale-data` files until we are ready to mass update them next year by simply running the `build-files` script again (with no `NEW_LOCALE`).

build-locale-data/build-files.js
```diff
@@ -3,6 +3,7 @@ import { generateLocaleData } from './generate-locale-data.js';
 
 const localeData = await generateLocaleData();
 
-await Promise.all([
-	buildIntlFiles(localeData)
-]);
+await buildIntlFiles(localeData);
+
+// TODO: either build more files for more repos, or remove if built elsewhere
```
dlockhart: Remove?

lib/locale-data/.DS_Store
dlockhart, line 1: Probably don't want this checked in?

lib/locale-data/current.js
```diff
@@ -1,17 +1,26 @@
...
 
 const documentLocaleSettings = getDocumentLocaleSettings();
 
+let currentLocale;
+
 async function setLocaleData() {
+	if (currentLocale === documentLocaleSettings.language) return;
 	const resolvedLanguage = resolveSupportedLocale(documentLocaleSettings.language);
 	Object.keys(localeData).forEach(k => delete localeData[k]);
+	//documentLocaleSettings._cache.delete('dateTimeDescriptor');
```
dlockhart: Remove?

lib/locale-data/locales/ca.js
```diff
@@ -30,7 +31,7 @@ export default {
 	dateFormats: {
 		full: "EEEE, d MMMM 'del' y",
 		long: "d MMMM 'del' y",
-		medium: "d MMM y",
+		medium: "d MMMM 'del' y",
```
dlockhart, line 36: So this is us _overriding_ what was in the CLDR in order to match our (wrong) existing data ya? And then we'll do some kind of mass fix in the future where we communicate with clients?

bearfriend: Yep that's right. Blog post planned for ~October about it all, then they'll have time to offer feedback etc. Then I think ~July we pull the trigger.

build-locale-data/build-files.js
bearfriend: Yeah, I guess so for now. Not like it's complicated to add back later.

lib/locale-data/locales/ca.js
bearfriend: For `medium` specifically, I don't think we were "wrong". It's just a mismatch in what we call things. Like this just matches `long` above it, as do a number of others. As you go back further to older locales it gets murkier.

lib/locale-data/current.js
```diff
@@ -1,17 +1,61 @@
...
+// all locales share the same keys, and consumers shouldn't be able to delete
+// properties, so block deletion while still allowing us to redefine values below
+export const localeData = new Proxy(localeDataTarget, {
+	deleteProperty() { return false; }
+});
 
 const documentLocaleSettings = getDocumentLocaleSettings();
 
+let currentLocale;
+let localeDataPromise = Promise.resolve();
```
bearfriend, line 16: A promise that can be overwritten each time the locale changes

lib/locale-data/current.js
```diff
@@ -1,17 +1,61 @@
...
 import { resolveSupportedLocale } from './supported.js';
 
-export const localeData = {};
+const localeDataTarget = {};
+
+// block deletions and existing property redefinitions
+export const localeData = new Proxy(localeDataTarget, {
+	deleteProperty() { return false; },
+	defineProperty() { return false; }
+});
```
bearfriend, line 11: Since we can't `Object.freeze` the shared object (so we can update its properties), we proxy it and disallow (re)defining and deleting properties outside this module.

lib/locale-data/current.js
```diff
@@ -1,17 +1,57 @@
...
 
 const documentLocaleSettings = getDocumentLocaleSettings();
 
+let currentLocale;
+let localeDataPromise = Promise.resolve();
+
+Object.defineProperty(localeDataTarget, 'then', { value: (...args) => localeDataPromise.then(...args) });
+Object.defineProperty(localeDataTarget, 'catch', { value: (...args) => localeDataPromise.catch(...args) });
+Object.defineProperty(localeDataTarget, 'finally', { value: (...args) => localeDataPromise.finally(...args) });
```
bearfriend, line 20: Make `localeDataTarget` thenable, always redirecting to the dynamic `localeDataPromise`. We use `defineProperty` here mostly to prevent enumerability. i.e. before first resolution, `Object.keys(localeData) // -> []`

This means `localeData` acts as its own promise, and allows existing and future intl functions like `formatDate` to remain synchronous, as long as there is a top-level `await localeData` anywhere before they can be called, which we already do right at the end of this module. Shifting that responsibility to consumers is an option that would prevent blocking while the initial locale data loads, but in practical terms would offer little to no benefit as something else is likely to be imported that already top-level awaits it. After that, any consumer is responsible for managing updates:

```javascript
// library
import { localeData } from '@brightspace-ui/intl/lib/locale-data/current.js';
import { resolveSupportedLocale } from '@brightspace-ui/intl/lib/locale-data/supported.js';

// current.js already awaited the initial locale's data
export function getMonths(format) {
	const months = localeData.months.standAlone[format];
}

export async function changeLanguage(lang) {
	document.documentElement.lang = resolveSupportedLocale(lang);
	await localeData;
}
```

```javascript
// component/mixin
...
import { localeData } from '@brightspace-ui/intl/lib/locale-data/current.js';

class MyComponent ... {
	...
	connectedCallback() {
		this.connectedCallback();
		this.addEventListener('d2l-locale-data-change', () => requestUpdate('localeData'));
	}

	render() {
		return html`
			<ul>${localeData.months.standAlone.wide.map(m => html`
				<li>${m}</li>`)}
			</ul>
		`;
	}
}
```

lib/locale-data/current.js
```diff
@@ -1,17 +1,54 @@
...
+	Object.keys(newData).forEach(k => localeDataTarget[k] = newData[k]);
+
+	currentLocale = documentLocaleSettings.language;
+	document.dispatchEvent(new CustomEvent('document-locale-data-change', { detail: { language: resolvedLanguage } }));
 }
 
-documentLocaleSettings.addChangeListener(setLocaleData);
+function updateLocaleData() {
+	localeDataPromise = setLocaleData();
+	return localeDataPromise;
```
bearfriend, line 35: Update the double proxied promise pointer

lib/locale-data/current.js
```diff
@@ -1,17 +1,54 @@
...
+			deepFreeze(value);
+		}
+	}
+	return Object.freeze(obj);
+}
 
-setLocaleData();
+registerLocaleDataListener();
+await updateLocaleData();
+Object.preventExtensions(localeDataTarget);
```
bearfriend, line 54: The proxy already prevents consumers adding new properties, but this ensures we aren't adding extraneous properties when we change locales. We can probably remove this in favor of more tests once generating/modifying the locale data files is more settled down, but for now it's useful to catch inconsistencies there.

lib/dateTime.js
```diff
@@ -692,300 +650,218 @@ export function convertUTCToLocalDateTime(date) {
...
 		};
 
 		if (settings.overrides.date) {
-			merge(descriptor, settings.overrides.date);
+			if (settings.overrides.date.calendar && settings.language.startsWith(resolveSupportedLocale(settings.language).toLowerCase())) {
+				// prefer more nuanced front-end data
+				delete settings.overrides.date.calendar.months;
+				delete settings.overrides.date.calendar.days;
+			}
+			return merge(JSON.parse(JSON.stringify(descriptor)), settings.overrides.date);
```
bearfriend, line 864: Prevents accidental/attempted overwriting of `localeData` subproperties since without this they would be direct references. We can't use `structuredClone` yet. Thanks Safari!

lib/locale-data/current.js
dlockhart: I agree the ability for a synchronous API is super nice!

lib/locale-data/current.js
```diff
@@ -2,25 +2,53 @@ import { getDocumentLocaleSettings } from '../common.js';
...
+	localeDataPromise = setLocaleData();
+	return localeDataPromise;
+}
+
 export function registerLocaleDataListener() {
-	documentLocaleSettings.addChangeListener(setLocaleData);
+	documentLocaleSettings.addChangeListener(updateLocaleData);
+}
+
+function deepFreeze(obj) {
```
dlockhart, line 42: 🥶

lib/common.js
```diff
@@ -83,7 +84,11 @@ function updateLocalNames() {
 		return;
 	}
 	supportedLocalesDetails.forEach(l => {
-		l.localName = localName.of(l.overrideCode || l.code);
+		try {
+			l.localName = localName.of(l.overrideCode);
```
dlockhart, line 88: Under what circumstances would using the override code throw? It seems like it's more "valid" than `code`?

dlockhart (approved):
Nice!

lib/common.js
bearfriend: The old override for `fr-ON`, `fr-CA-Ontario`, actually got us `Canadian French (ONTARIO)`, which is a little odd but pretty good. The new override, `fr-CA-u-sd-caon` is more correct and lets us figure out more for the `localeData`, but apparently `Intl.DisplayNames` doesn't like it, so we fall back to `fr-ON` and get `French (ON)` now. Not ideal, but I think this only used in the demo right now and it can be improved with the actual `localeData`.

PR #388, opened 2026-07-09
Only overwrite languageId when requested

bearfriend:
[GAUD-10152](https://desire2learn.atlassian.net/browse/GAUD-10152): OSLO > Custom languages are not resolved via oslo - falls back to parent language

In order for OSLO to pull down custom langpacks that use (via a custom locale) locale codes that we already support, we need to limit overwriting the `languageId`. Currently we always check if there is a mismatch between the document language and the OSLO `languageId`. With a custom langpack/locale, this mismatch could be intentional.

[GAUD-10152]: https://desire2learn.atlassian.net/browse/GAUD-10152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/localize.js
+++ b/lib/localize.js
@@ -238,8 +238,9 @@ export const getLocalizeClass = (superclass = class {}) => class LocalizeClass e
 				if (osloCollection) {
 					const oslo = this.documentLocaleSettings.oslo;
 					let languageId = 0;
-					if (oslo.collection) {
-						languageId = supportedLocalesDetails.find(l => l.code === osloLang || l.pack === osloLang)?.id;
+					const doc = document.documentElement;
+					if (doc.hasAttribute('data-custom-lang') && oslo.collection) {
+						languageId = Number(doc.dataset.languageId) || supportedLocalesDetails.find(l => l.code === osloLang || l.pack === osloLang)?.id;
 						if (languageId) {
 							const batch = oslo.batch && new URL(oslo.batch, self.origin);
 							if (batch && languageId !== batch.searchParams.get('languageId')) {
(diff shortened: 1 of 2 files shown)


lib/localize.js
```diff
@@ -238,8 +238,9 @@ export const getLocalizeClass = (superclass = class {}) => class LocalizeClass e
 				if (osloCollection) {
 					const oslo = this.documentLocaleSettings.oslo;
 					let languageId = 0;
-					if (oslo.collection) {
-						languageId = supportedLocalesDetails.find(l => l.code === osloLang || l.pack === osloLang)?.id;
+					const doc = document.documentElement;
+					if (doc.hasAttribute('data-custom-lang') && oslo.collection) {
```
bearfriend, line 242: The attribute will come from https://github.com/Brightspace/lms/pull/89913

lib/localize.js
```diff
@@ -238,8 +238,9 @@ export const getLocalizeClass = (superclass = class {}) => class LocalizeClass e
...
+						languageId = Number(doc.dataset.languageId) || supportedLocalesDetails.find(l => l.code === osloLang || l.pack === osloLang)?.id;
```
bearfriend, line 243: This allows content authors to declare a preferred (custom) `languageId` that will be respected regardless of any locale code confusion

lib/localize.js
dlockhart: I'm confused... where do `data-custom-lang` and `data-language-id` get set? That linked PR is just emptying out `data-custom-lang` in content pages.

bearfriend: It's a boolean. Empty string sets the attribute. `delete` is what removes it.

bearfriend: And `data-language-id` would be set by an author on their content in edge cases.

dlockhart: We talked about this offline -- on the same page now.

dlockhart (approved):

PR #396, opened 2026-07-28
Add skeleton support to existing date/time functions

bearfriend:
[GAUD-10218](https://desire2learn.atlassian.net/browse/GAUD-10218): Update intl dateTime functions to accept and render from skeleton patterns

For now, we only call `formatDateTimeSkeleton` when the `skeleton` option is explicitly set. Down the road we can start to transition more calls/options to use/convert to skeletons and slowly get rid of the existing custom format handling.

[GAUD-10218]: https://desire2learn.atlassian.net/browse/GAUD-10218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/dateTimeSkeleton.js
+++ b/lib/dateTimeSkeleton.js
@@ -1,0 +1,78 @@
+import { getDocumentLocaleSettings } from './common.js';
+import IntlMessageFormat from 'intl-messageformat';
+import { localeData } from './locale-data/current.js';
+
+const documentLocaleSettings = getDocumentLocaleSettings();
+
+export function formatDateTimeSkeleton(dateTime, options = {}) {
+	const { date = true, time = true, locale } = arguments[2] || {};
+
+	const {
+		timezone: timeZone =
+			options.timeZone
+			|| documentLocaleSettings.timezone.identifier
+			|| undefined,
+		timeZoneName,
+		hourCycle
+	} = options;
+	let {
+		format,
+		skeleton = '',
+	} = options;
+
+	if (!skeleton) {
+		if (date) {
+			if (!format || format === 'short') {
+				skeleton = localeData.dateFormats.short;
+			}
+			else if (format === 'medium') {
+				skeleton = localeData.dateFormats.medium;
+			}
+			else if (format === 'full') {
+				skeleton = localeData.dateFormats.full;
+			}
+		}
+		if (time) {
+			skeleton += localeData.timeFormats.short;
+		}
+
+		if (!skeleton) {
+			console.warn(`"${format}" is not a valid format`);
+			return;
+		}
+
+		format = null;
+	} else {
+		if (!options.forceUnsupportedFormat) {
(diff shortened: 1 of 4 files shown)


lib/dateTimeSkeleton.js
```diff
@@ -0,0 +1,78 @@
...
+		}
+	};
+	const dateFormatter = new IntlMessageFormat(msg, locale || documentLocaleSettings.language, null, { formatters });
+	const formatted = dateFormatter.format({ dateTime });
+	return formatted;
+}
+
+const customFormatMap = new Map([
+	['MM/dd/yy', 'en-US', 'MMddyy']
+]);
```
bearfriend: Whoops, didn't mean to commit this part.
```suggestion
```

test/dateTime.test.js
```diff
@@ -990,6 +997,67 @@ describe('dateTime', () => {
...
+			const result = formatDateTimeSkeleton(date, { skeleton, forceUnsupportedFormat: true }, { time: false });
+			expect(result).to.equal('2021');
+		});
+		it('should ignore date symbols when `date` is false', () => {
+			const date = new Date(2021, 0, 15, 14, 30);
+			const skeleton = 'GyMLdEecHH';
+			const result = formatDateTimeSkeleton(date, { skeleton, forceUnsupportedFormat: true }, { date: false });
+			expect(result).to.equal('2 PM');
+		});
+		it('should render stand-along months appropriately', async() => {
```
dlockhart:
```suggestion
		it('should render stand-alone months appropriately', async() => {
```

test/dateTime.test.js
```diff
@@ -990,6 +997,67 @@ describe('dateTime', () => {
...
+			await setLanguage('ca-ES');
+			const date = new Date(2021, 3, 15, 14, 30);
+			let skeleton = 'MMMM';
+			let result = formatDateTimeSkeleton(date, { skeleton, forceUnsupportedFormat: true });
+			expect(result).to.equal('abril');
```
dlockhart, line 1048: Bringing back nightmares...

lib/dateTimeSkeleton.js
```diff
@@ -0,0 +1,74 @@
...
+			|| undefined,
+		timeZoneName,
+		hourCycle
+	} = options;
+	let {
+		format,
+		skeleton = '',
+	} = options;
+
+	if (!skeleton) {
```
dlockhart, line 23: I'm a bit confused, since I don't see how this branch would be reached unless someone called `formatDateTimeSkeleton` directly without going through `formatDate` or `formatTime`? Is the plan to switch everything to this eventually and then retire `formatDate` and `formatTime`?

bearfriend: Yep exactly. Definitely some future-looking stuff there.

dlockhart (approved):

PR #406, opened 2026-08-24
FIx / improve running build-locale-data scripts locally

boarnoah:
Adding support for new locales involve running scripts ``node /build-locale-data/build-files.js`` 

But that fails because it is missing ``cldr`` dependency + node not having the document DOM API.

I am not sure how @bearfriend typically runs this script so I might have misunderstood. The script is running file locally now, but it generates a lot of locale-data changes?

PS: I am associating this with the vulcan card to add Gaelic support, but this might be more appropriate under a Gaudi card re: improvements in this area? 

### DIFF
--- a/lib/common.js
+++ b/lib/common.js
@@ -34,37 +34,6 @@ export function getLanguage() {
 	return langTag;
 }
 
-export function merge(obj1, obj2, keepOriginal = false) {
-	if (obj2 === undefined || obj2 === null || typeof(obj2) !== 'object') {
-		return;
-	}
-	for (const i in obj2) {
-		// eslint-disable-next-line no-prototype-builtins
-		if (obj1.hasOwnProperty(i)) {
-			if (typeof(obj2[i]) === 'object' && typeof(obj1[i]) === 'object') {
-				merge(obj1[i], obj2[i]);
-			} else {
-				if (keepOriginal) obj1[`_original_${i}`] = obj1[i];
-				obj1[i] = obj2[i];
-			}
-		}
-	}
-	return obj1;
-}
-
-export function validateFormatValue(value) {
-	if (value === undefined || value === null) {
-		return 0;
-	}
-	if (typeof value === 'string') {
-		value = parseFloat(value);
-	}
-	if (isNaN(value) || typeof value !== 'number') {
-		throw new RangeError('value is out of range.');
-	}
-	return value;
-}
-
 let documentLocaleSettings = null;
 export function getDocumentLocaleSettings() {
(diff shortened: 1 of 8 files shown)


lib/common.js
```diff
@@ -34,37 +34,6 @@ export function getLanguage() {
 	return langTag;
 }
 
-export function merge(obj1, obj2, keepOriginal = false) {
```
boarnoah, line 37: this is used in build-locale-data scripts, but failing like so:

```
> node ./build-locale-data/build-files.js

file:///C:/D2L/Code/intl/lib/document-locale-settings.js:1
const host = document.documentElement;
             ^

ReferenceError: document is not defined
    at file:///C:/D2L/Code/intl/lib/document-locale-settings.js:1:14
    at ModuleJob.run (node:internal/modules/esm/module_job:439:25)
    at async node:internal/modules/esm/loader:643:26
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)
```

so refactored it out, since they are more standalone util functions not dependent on DOM APIs

package.json
```diff
@@ -1,14 +1,15 @@
...
   "description": "Internationalization APIs for number, date, time and file size formatting and parsing in D2L Brightspace.",
   "main": "lib/number.js",
   "type": "module",
   "scripts": {
     "lint": "eslint .",
     "start": "web-dev-server --node-resolve --watch --open /demo/",
     "test:unit": "d2l-test-runner",
-    "test": "npm run lint && npm run test:unit"
+    "test": "npm run lint && npm run test:unit",
+    "build-locale-data": "node ./build-locale-data/build-files.js"
```
boarnoah, line 12: assuming there is a reason this wasn't already done this way? In which case, is a separate package.json for the build-locale-data more appropriate?

package.json
```diff
@@ -41,6 +42,7 @@
   "homepage": "https://github.com/BrightspaceUI/intl",
   "devDependencies": {
     "@brightspace-ui/testing": "^1",
+    "cldr": "^8.0.0",
```
boarnoah, line 45: assumed latest is fine*? I do think this kind of stuff needs to be version controlled so anyone could run these locale build scripts?


\* It might be the cause for these changes I see... was I supposed to use a different version?

dlockhart:
Thanks @boarnoah -- I've converted this into a draft for now so we can wait for @bearfriend to have a look. I'm not sure if everything is in a complete state or if there was more to do.

dlockhart:
I'm going to close this for now to get it off our "to review" list, but something Danny can review when he's back.

PR #408, opened 2026-08-24
Add Gaelic (Irish) support

boarnoah:
I added a row to lib/locale-data/support.js like:

```js
{ id: 42, code: 'ga', source: 'ga', pack: 'ga', dir: 'ltr', name: 'Gaeilge (Éire)' },
```
based on https://www.localeplanet.com/icu/ga/index.html

then ran ``/build-locale-data/build-files.js``

which in turn changed the row to what it is now + generated the ``lib/locale-data/locales/ga.js`` as well as a whole lotta other changes which I didn't include.

It's highly likely I am using this script wrong locally

### DIFF
--- a/lib/locale-data/locales/ga.js
+++ b/lib/locale-data/locales/ga.js
@@ -1,0 +1,391 @@
+export default {
+	languageDisplayName: "Gaeilge",
+	localeDisplayName: "Gaeilge",
+	sourceLocale: "ga",
+	localeCode: "ga",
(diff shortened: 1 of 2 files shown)


boarnoah:
Talked to @dlockhart . 

Stacked these mostly so someone else could follow along re: how I generated these file contents.

But we want Danny's go ahead before we make changes to the scripts etc... so for now we will try to just get these generated Gaelic things published by itself first.

---

Going to close and re-open since there is no way to de-stack

PR #410, opened 2026-08-25
Add Gaelic (Irish) support

boarnoah:
This is re-opening #408 as a standalone PR

---

I added a row to lib/locale-data/support.js like:

```js
{ id: 42, code: 'ga', source: 'ga', pack: 'ga', dir: 'ltr', name: 'Gaeilge (Éire)' },
```
based on https://www.localeplanet.com/icu/ga/index.html

then ran ``/build-locale-data/build-files.js``

which in turn changed the row to what it is now + generated the ``lib/locale-data/locales/ga.js`` as well as a whole lotta other changes which I didn't include.

It's highly likely I am using this script wrong locally

### DIFF
--- a/lib/locale-data/locales/ga.js
+++ b/lib/locale-data/locales/ga.js
@@ -1,0 +1,391 @@
+export default {
+	languageDisplayName: "Gaeilge",
+	localeDisplayName: "Gaeilge",
+	sourceLocale: "ga",
+	localeCode: "ga",
+	likelySubtag: "ga-Latn-IE",
+	layout: {
+		orientation: {
+			characterOrder: "left-to-right",
+			lineOrder: "top-to-bottom"
+		}
+	},
+	pluralClass: {
+		cardinal: [
+			"one",
+			"two",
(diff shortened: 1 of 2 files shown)


lib/locale-data/locales/ga.js
boarnoah, line 1: I can't attest to the validity of this. 

In particular, a little wary about the fact that the script decided to rename ``Gaelic (Irish)`` to just be ``Gaelic`` - I doubt this will be a issue unless try to support Gaelic (Scottish) as well but still ☹️

dlockhart: I think it's OK. All our other display names just have the language written in that particular language -- with a few exceptions where we display territorial names inside (e.g. English (Canada)).

The Scottish variant would be a different locale code (`gd`) and would translate as `Gàidhlig`.

dlockhart (approved):
This looks good, although admittedly Danny's the expert here. But I think we can correct any issues he finds later to get you unblocked.

